PR #7, opened 2019-08-06
Heads up Notification for Android 7+

goldjay:
* Android 7+ requires a message channel
* Channel name matches the channel we are using on the backend to send notifications

Next Steps:
* Lower-end Android phones (5+) still not displaying heads-up notifications
* Our notifications meet the three requirements that **may** get heads-up notifications to work:
full-screen intent, high priority and sound


EbPhotoStore/src/ui/AppController.tsx
```diff
@@ -87,6 +102,8 @@ class AppController extends React.Component<Props, State> {
     }
 
     AppState.removeEventListener("change", this._handleAppStateChange);
+
+    firebase.notifications().android.deleteChannel("eb_photo_store");
```
audererob: Does anything need to be done with this._channel here, such as setting it to null?

goldjay: @audererob I think I wouldn't need to do anything since the component would be unmounted.

PR #10, opened 2019-08-06
Record whether the flash was on when RDT picture was taken

balderdash:


balderdash:
>There are other methods that set fields of RDTInfo rather than using a generalized one - e.g. setTotalTestStripTime, setRDTCaptureTime, setRDTReaderResult, setRDTInterpretationShown, and setResultShown. It seems that perhaps we should either have setRDTPhotoFlashEnabled or convert those other methods over to use the new generalized one. Just a thought.

I made the redux action more specific, rather than a general `rdtInfo` updater, but kept the name since I also added another field "flashDisabledAutomatically".

PR #15, opened 2019-08-07
Requiring authentication to access Firestore

billy-audere:
Continuation of https://github.com/AudereNow/learn/pull/1778 with some additional testing and an update to the app to pass the required health worker uid from auth.


EbPhotoSite/firestore.rules
```diff
@@ -1,7 +1,82 @@
 service cloud.firestore {
   match /databases/{database}/documents {
+    function getRole() {
+      return get(/databases/$(database)/documents/users/$(request.auth.uid)).data.role;
+    }
+
+    function hasRole() {
+      return exists(/databases/$(database)/documents/users/$(request.auth.uid));
+    }
```
rightparen: Roles are generally many-to-many.  Consider:
```
function hasRole(role) {
  const user = /databases/$(database)/documents/users/$(request.auth.uid);
  return exists(user) && (get(user).data.roles.indexOf(role) >= 0);
}
```

Then the check below would be:
```
  allow read, write: if hasRole("admin");
```

billy-audere: const, if/else, indexOf, etc are unsupported so the rule is going to be a little different.  I'm just going to create a document per role instead of a document per user.

billy-audere:
I will be waiting to merge this since the health worker uid isn't set on current data & it'll probably be best to wipe data in staging and re-load production.

PR #18, opened 2019-08-07
Fix data pipeline update causing nginx to timeout

rightparen:
There are cases where updating the data pipeline takes ~90 seconds.  It appears this is the cause of the toggling alarm for cough import (since we update the data pipeline immediately after import).

This propagates the trick of writing keep-alive whitespace throughout the update.

Because we don't return any meaningful data from a pipeline update, we may also just want to kick off the update asynchronously and ignore the result, but I'd like to discuss tradeoffs around being able to alarm on that.  Another alternative is to use a SQL LIMIT to keep queries granular, but that seems more complex.


rightparen:
This is https://auderenow.atlassian.net/browse/FEV-1042

billy-audere:
Would it be worth solving this by setting proxy_read_timeout in the nginx configuration?

Longer term I would prefer separate steps for loading raw data from cascading downstream & also not having to invoke these through an HTTP endpoint.

rightparen:
@billy-audere I had tried increasing timeouts earlier for the import, but we proxy through several layers and it was problematic to get node, Express, nginx, the ELB, and maybe others to agree on a multiple-minute timeout for specific paths and to debug which layer had issues when something went wrong.

Agreed that these should eventually not be invoked through an HTTP endpoint, or at least that the HTTP request should not wait while the process runs.

I'm not sure I understand the issue re: separating loading and cascading.  Import and cascade are already implemented separately at this point--the reason for triggering them close in time is that the import is one of the main sources of modified state that needs to be propagated.

billy-audere:
@rightparen We don't have concurrency control on refreshing derived tables since it is part of individual composite processes that run & failures surface at a coarser level.

We've talked a bit about DAG solutions but something less involved could be pushing a messages to an SNS topic to queue a derived schema refresh via the updateDerived endpoint.  It isn't top of mind but fits my personal preferences if we see an increase in the number of data sources we are consuming or more complexity in our data pipelines.

PR #21, opened 2019-08-07
Move prettier command line arg to config file

balderdash:
I also run `prettier` in my editor. This way both pre-commit and vim run it with the same config.


tpaik18:
@balderdash If Prettier up til now has removed the trailing comma, then can we fix the .pre-commit-config.yaml file to just do trailingComma = none?

balderdash:
@tpaik18 When Jenny added precommit in https://github.com/AudereNow/learn/pull/1794, she added all the trailing commas to our codebase.

tpaik18:
Oh! Okay then this is cool~

PR #23, opened 2019-08-07
FEV-1066 Indicate patients that have a new message available in the Patient List view

audererob:
[FEV-1066](https://auderenow.atlassian.net/browse/FEV-1066)
We'll remember the last viewed time of messages, and only show the new message icon for a patient in the patient list view if a message is seen with a newer timestamp.


EbPhotoStore/src/ui/Details.tsx
```diff
@@ -110,6 +111,10 @@ class Details extends React.Component<Props & WithNamespaces, State> {
     this._phoneInput = React.createRef<NumberInput>();
     this._detailsInput = React.createRef<TextInput>();
     this._notesInput = React.createRef<TextInput>();
+
+    if (!this.props.isNew) {
+      this.props.dispatch(resetMessageLastViewedAt(this.props.id));
```
balderdash: You'll also want to update the timestamp if a new message comes in when they already have the details page open (perhaps only if they've scrolled down to the chat area, but that can come later).

audererob: Good call, making that change now

audererob: I wrote the code to look at when the messages array changes (componentDidUpdate) but then released I could cheat a bit and just put the timestamp reset in the "nav back" handling.

PR #32, opened 2019-08-07
Make patients with unread updates appear in bold on the patient list page

balderdash:
This compares the triage document's `lastViewed`, the encounter document's `updatedAt`, and the most recent message `timestamp` to determine which patients have new changes or messages that have not yet been reviewed by an admin.

I added a firebase listener per encounter, rather than one big collection group query because I couldn't figure out how to write a query that returned just one message per patient.

image


balderdash:
Completed in https://github.com/AudereNow/audere/pull/54

PR #46, opened 2019-08-08
RDT Reader updates from CJ

balderdash:
CJ's Changelog:

```
Android
- optimization for release env. remove unnecessary debug logs and refactoring.
- do all the quality checks for every frame
- further reduce the size of image for sift operations. computation/frame is reduced from 250ms to 150ms on pixel 1
- Add croppedRDTMat to the callback or storing the perspective corrected test strip image in the future
iOS
- do all the quality checks for every frame
- Add croppedTestStripImage to the callback for storing the perspective corrected test strip image in the future
```

We're not actually saving the cropped RDT image yet, we'll add that for the next release.

I also added a `testStripDetected` boolean to the callback, which I'll be using to change the way we automatically turn off the flash.


balderdash:
@audererob No longer creating UIImages unless the checks passed. Looks like they already handled it that way on android: https://github.com/AudereNow/audere/pull/46/files#diff-959213aaa4304f4b355b787d2b5f1f53R267-R269

PR #50, opened 2019-08-09
Subscribe to individual triage documents instead of collections

billy-audere:
Also setting permissions on messages collections.


billy-audere:
Subscribing to the entire collection for triage updates caused a permission error since we were trying to receive updates about all documents.

PR #53, opened 2019-08-12
FEV-1078 Implemented PhotoUploader in native

Ninjestic:


EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/PhotoUploader.java
```diff
@@ -0,0 +1,269 @@
...
+  private final String TAG = "PhotoUploader";
+
+  private final int second = 1000;
+  private final int minute = 60 * second;
+  private final int retryDelay = 20 * minute;
+
+  private final String pendingDir;
+  private final String pendingDirPrefix;
+
+  private final String firestoreDocID = "kcn6PxLyXgo2POMeuohB";
```
goldjay: Is this leftover from testing?

EbPhotoStoreNative/app/build.gradle
```diff
@@ -1,14 +1,16 @@
 apply plugin: 'com.android.application'
+apply plugin: 'io.fabric'
 
 android {
     compileSdkVersion 28
     defaultConfig {
-        applicationId "com.example.ebphotostorenative"
-        minSdkVersion 15
+        applicationId "src.main.java.com.example.ebphotostorenative"
```
audererob: Does this need to change as part of the upload functionality? Longer term I think we should make it org.auderenow.ebphotostorenative but was thinking that can happen whenever.

EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/PhotoUploader.java
```diff
@@ -0,0 +1,269 @@
...
+
+  public PhotoUploader(Activity context) {
+    this.pendingEvents = new ArrayList<Event>();
+    this.failedFiles = new TreeSet<String>();
```
goldjay: We may not need a TreeSet since we are just checking if a file exists in failedFiles.

EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/PhotoUploader.java
```diff
@@ -0,0 +1,323 @@
...
+    String summary = func + ": " + location + " threw Error: '" + message + "' due to: " + cause;
+    Bundle bundle = new Bundle();
+    bundle.putString("error_message", message);
+    bundle.putString("cause", cause);
+    bundle.putString("location", location != null ? location : "");
+    mFirebaseAnalytics.logEvent("Error", bundle);
+    return new LoggedError(summary);
+  }
+
+  private static class LoggedError extends Error {
```
rightparen: This should extend RuntimeException or Exception if you want it checked.  Error is for conditions you shouldn't catch.

EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/PhotoUploader.java
```diff
@@ -0,0 +1,323 @@
...
+    Event(String type) {
+      this.type = type;
+    }
+
+    Event(String type, String photoId, String jpegBase64) {
+      this.type = type;
+      this.photoId = photoId;
+      this.jpegBase64 = jpegBase64;
+    }
+  }
```
rightparen: The reason for having events in the other implementation was because there was an array and a single `Runnable`.  Since you are using separate `Runnable`s for   It does seem useful to have the paired methods `xyz` and `looperHandleXyz`, but at this point you could probably remove Event and just have each `xyz` method post its own `Runnable`.  Then things like `photoId` and `jpegBase64` are implicitly captured and you can just pass them as arguments via the `Runnable`.

EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/PhotoUploader.java
```diff
@@ -0,0 +1,323 @@
...
+      while ((c = inputStreamReader.read()) != -1) {
+        outputStreamWriter.write(c);
+      }
+      inputStreamReader.close();
+      outputStreamWriter.close();
+      filePath.delete();
+    } catch (IOException err) {
+      logError("looperUploadSuccess", "PhotoUploader", new Error("Failed to save"));
+      throw new Error("There was a problem writing " + filePath.toString());
+    }
```
rightparen: This block can be replaced with https://developer.android.com/reference/java/nio/file/Files.html#move(java.nio.file.Path,%20java.nio.file.Path,%20java.nio.file.CopyOption...)

EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/PhotoUploader.java
```diff
@@ -0,0 +1,323 @@
...
+          looperUploadSuccess(moveOnSuccess);
+        }
+      }).addOnFailureListener(new OnFailureListener() {
+        @Override
+        public void onFailure(Exception e) {
+          Log.w(TAG, "Error writing document", e);
+          looperUploadFail(filePathFinal);
+        }
+      });
+    } catch (Error e) {
```
rightparen: Never catch Error.

You should just catch the IOException from above here.  If Firebase throws on errors (instead of just calling onFailure), you should catch those too.

In the error case, you should add the file to the error list and try the next one.

EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/PhotoUploader.java
Ninjestic: I was using it to firebaseAnalytics log anything that went wrong but I guess I should leave that to crashlytics?

PR #57, opened 2019-08-12
Add Save button to details page

jennya:
Request for feedback -- not sure if this is exactly how we want to do this, just a first pass.

Current behavior:
* Save button that is only enabled when there are unsaved changes
* Alert that notifies users when they hit the back button with unsaved changes (language needs work)
* Hitting the add photo button saves any unsaved changes as before

Screenshot_20190812-111125
Screenshot_20190812-111132
Screenshot_20190812-111137


tpaik18:
"Continue to discard changes" confuses me. Also not sure what Cancel means. How about

There are unsaved changes 

Save    Discard

jennya:
@tpaik18 sounds good to me

jennya:
Latest iteration:

Screenshot_20190813-095746

ShawnaCooper:
This doesn't need to be handled immediately - but the spacing of that
dialog seems a bit off to me.
@David Schacht <david@auderenow.org> - could you look on Jenny's machine
please and see if we can make an adjustment?

Thanks!
-Shawna

PR #61, opened 2019-08-12
FEV-950 Record whether the flash was on when the RDT image was captured

audererob:
[FEV-950](https://auderenow.atlassian.net/browse/FEV-950)
Factor in whether flash/torch mode is supported on the device.


audererob:
@tpaik18 If build 90 is being submitted to the store then this should just go into master, correct?

PR #65, opened 2019-08-13
FEV-1068 Retry fetching photo urls on details page

balderdash:


balderdash:
They only fail for a very brief period after the photo gets uploaded, but that becomes an issue with the details page showing updates in real time.

Of course they also fail if the photo hasn't uploaded yet, but that shouldn't happen after https://github.com/AudereNow/audere/pull/64 goes in.

PR #68, opened 2019-08-14
Pending Photo Upload Indicator

goldjay:
[FEV-1086](https://auderenow.atlassian.net/browse/FEV-1086)

* Checks for status: non-uploaded photos in the store
* photoID is randomly-generated and it's possible to have the same id for two different photos so I'm hesitant to remove uploaded photos. Do you think that is something to be concerned about?

Screen Shot 2019-08-14 at 1 20 27 PM


balderdash:
I wouldn't worry about the photo uuids colliding.

EbPhotoStore/src/ui/components/TitleBar.tsx
```diff
@@ -83,9 +93,19 @@ class TitleBar extends React.Component<Props & WithNamespaces> {
 
 export default connect((state: StoreState) => ({
   demoMode: state.meta.demoMode,
+  hasPendingPhoto: state.meta.hasPendingPhoto,
```
balderdash: Instead of saving the pending photos state separately under meta, can we call `hasPendingPhotos(state)` here? It's just nicer to have a single source of truth.

If performance does turn out to be a problem, we can remove completed photo uploads from the photo uploads state.

goldjay: @balderdash Nice! I'll make the changes

PR #79, opened 2019-08-16
Copyright

Jaystar2323:
FEV-558 Adding license headers


scripts/update_copyright.py
```diff
@@ -1,18 +1,37 @@
...
-    text.insert(0, header_text)
+    try:
+        text = f.readlines()
+    except:
+        sys.stderr.write("Could not read file")
+        return
+
+    header_found = False
+    for i in range(5 if len(text)>=5 else len(text)):
+        (new_str, num) = re.subn(r"\s?Copyright\s\(c\)\s+\d{4}(-\d{4})? by Audere", header_text[0], text[i])
```
auderephilip: [minor] Please try to format lines to 80 chars or less when possible.  Not positive in python that it is... but if so, please do that.

[minor] We might consider making the last part of the regex `(\s*-\s*\d{4})?` instead, or something like that, to allow for spaces between the two years in a range.

[more important] This captures all copyrights well — great work!  The only remaining case we need to handle is where the existing copyright already has a year range, in which case we just want to update the last year to the current year.  So we'd want "2016-2018" to become "2016-2019", instead of just become "2019".

scripts/update_copyright.py
```diff
@@ -1,18 +1,37 @@
 import os
 import argparse
+import sys
+import re
 
-header_text = "// Copyright (c) 2019 by Audere\n//\n// Use of this source code is governed by an MIT-style license that\n// can be found in the LICENSE file distributed with this file.\n"
+header_text = [" Copyright (c) 2019 by Audere", "", " Use of this source code is governed by an MIT-style license that", " can be found in the LICENSE file distributed with this file."]
```
auderephilip: Great work splitting this up.  Will have other benefits too.

Could we make `2019` not hard-coded, but instead take from the year of today's date?

scripts/update_copyright.py
```diff
@@ -1,18 +1,37 @@
...
+    if not header_found:
+        print("Create new header")
+
+        if(text[0].startswith("#!")):
+            text.insert(1, header_text[0])
+        else:
+            text.insert(0, header_text[0])
     f.close()
 
     f = open("test.txt", "w")
```
auderephilip: Let's only write the output file if the input file should actually change.  In other words, only write the file if either a header wasn't found, or if an outdated header was found.  In cases where we've found a header with valid dates already, we shouldn't write the whole file back out.

scripts/update_copyright.py
```diff
@@ -20,11 +24,25 @@ def editFile(path):
 
     header_found = False
     for i in range(5 if len(text)>=5 else len(text)):
```
auderephilip: [minor] Suggest pulling "5" out into a constant near the top of the file.  In general, giving constants names will help reduce future errors (by other coders, for instance) when people come back in to edit this file.

scripts/update_copyright.py
```diff
@@ -20,11 +24,25 @@ def editFile(path):
...
-        (new_str, num) = re.subn(r"\s?Copyright\s\(c\)\s+\d{4}(-\d{4})? by Audere", header_text[0], text[i])
-        if num > 0:
-            text[i] = new_str
-            header_found = True
-            break
+        try:
+            old_str = re.search(r"\s?Copyright\s\(c\)\s+\d{4}(\s*-\s*\d{4})? by Audere",
+                 text[i]).string
+        except:
+            print("no match found")
```
auderephilip: I suspect you'll likely want to remove this (since it'll print for each line that doesn't match, right?).

scripts/update_copyright.py
```diff
@@ -20,11 +24,25 @@ def editFile(path):
...
+            continue
+        if str(now.year) in old_str:
+            f.close()
```
auderephilip: Given that `f.close()` now needs to happen in multiple places (e.g. here and in line 53), I think it'd likely be cleaner to close the file higher up in the function (right after you've read all the lines) because you don't need the file for read anywhere after that anyway.  Then others won't need to be super-careful to remember to close the file if they add another return somewhere...

scripts/update_copyright.py
```diff
@@ -20,11 +24,25 @@ def editFile(path):
 
     header_found = False
     for i in range(5 if len(text)>=5 else len(text)):
-        (new_str, num) = re.subn(r"\s?Copyright\s\(c\)\s+\d{4}(-\d{4})? by Audere", header_text[0], text[i])
-        if num > 0:
-            text[i] = new_str
-            header_found = True
-            break
+        try:
+            old_str = re.search(r"\s?Copyright\s\(c\)\s+\d{4}(\s*-\s*\d{4})? by Audere",
```
auderephilip: A few suggested edits here:
- I suspect the start of the search should perhaps be `\W+`, not `\s?`, since the latter could technically match "weirdCopyright" etc (IIUC).
- It might be more flexible to allow `\s+` in front of the (c)

Bigger question:  I thought `re.search()` returns a `match` object or `None`, according to python docs, not a string.  Are you able to use `old_str` as a string?  And is the catch necessary below, given that you can test the output of `re.search` against `None`?

scripts/update_copyright.py
```diff
@@ -20,11 +24,25 @@ def editFile(path):
...
+                 text[i]).string
+        except:
+            print("no match found")
+            continue
+        if str(now.year) in old_str:
```
auderephilip: (not positive how this works, given above comment about `re.search`'s return value... LMK what you find...)

scripts/update_copyright.py
```diff
@@ -20,11 +24,25 @@ def editFile(path):
...
+            f.close()
+            print("already current year")
+            return
+        years = re.findall(r"\d{4}", old_str)
```
auderephilip: If `re.search` does indeed return a `match` object, it'll contain "match groups" (see https://docs.python.org/3.4/library/re.html#match-objects), which would make this code unnecessary (because it'd already tell you where the matches are).  The approach here isn't wrong;  just mentioning an observation that might make the code a little more efficient, and use regex matching in the ideal way (where you ask for match groups and you process the groups).

The match groups also tell you the start/end of the match, which allows the string substitution to potentially not need regex below either.

Not required to change this -- but I thought I'd mention it in case it leads to a new way of thinking about it, or cleaner code.

scripts/update_copyright.py
Jaystar2323: It does return a match object, but there is a `.string` that allows you to use it as a string

auderephilip: @Jaystar2323 : Ah, that makes sense.  Which is why you have the try/catch;  now I get it.

I think instead it might be cleaner to first just get a match object or  `None` back from `search()`, and then only use `.string` if it's not `None`, in which case you don't need the try/catch.  Exceptions can often be expensive when running code (performance-wise), and this current design would throw exceptions on most lines (every line which doesn't match the regex).

scripts/update_copyright.py
```diff
@@ -21,36 +22,28 @@ def editFile(path):
...
-
+    f.close()
     header_found = False
-    for i in range(5 if len(text)>=5 else len(text)):
-        try:
-            old_str = re.search(r"\s?Copyright\s\(c\)\s+\d{4}(\s*-\s*\d{4})? by Audere",
-                 text[i]).string
-        except:
-            print("no match found")
+    for i in range(5 if len(text)>=search_length else len(text)):
```
auderephilip: [minor] would `range(search_length if len(text)>=search_length`... work?

PR #80, opened 2019-08-16
Adds WhatsApp integration for EVD Track

auderephilip:
This introduces two Cloud Functions from us:  one to intercept all WhatsApp messages to EVD Track's bot, and one to process Dialogflow's AI-driven conversation with the user.

We add two node modules:  dialogflow and dialogflow-fulfillment.

Photo storage isn't implemented yet, but is easy enough in a follow-on diff.  We store CHW information in the "whatsAppCHWs" collection so that we don't pollute EVD Track's mainline collections;  we store patient info in "whatsAppPatients" for similar reasons.  In a future diff, we can copy the relevant CHW information and patient information into the official EVD Track collections when we're ready for full integration.

IMG_0B7786DB1C61-1


auderephilip:
@balderdash : Intents and the general AI training get configured on Google's Dialogflow console, which is attached to the same permissions as our Firebase staging app:  https://console.dialogflow.com/api-client/#/agent/ee37aa45-b2f6-4002-9ae5-00e321b544c7/intents.  The whole thing is remarkably easy to change... and when you do, they rebuild the model real-time.

PR #83, opened 2019-08-19
Native Geolocation

goldjay:
* NOTE: I had to change a few things in the photoUploader because our min SDK is 16 and the changed functions required API-26


EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/LocationFinder.java
```diff
@@ -0,0 +1,161 @@
...
+            isNetworkEnabled = locationManager
+                    .isProviderEnabled(LocationManager.NETWORK_PROVIDER);
+            if (!isGPSEnabled && !isNetworkEnabled) {
+                // no network provider is enabled
+                 Log.e("Network-GPS", "Disable");
+            } else if(checkPermission(context)) {
+                this.canGetLocation = true;
+                // First get location from Network Provider
+                if (isNetworkEnabled) {
+                    locationManager.requestLocationUpdates(
```
audererob: Is requestLocationUpdates required if we only want getLastKnownLocation? If so, should we turn it off after some period?

goldjay: @audererob That's a good point. I'll add something to remove updates after we get the location. It's not required if we only want to getLastKnowLocation, but I wanted to give both options and default to network-based location if we have permissions since it doesn't drain battery life like GPS-based location. If you think we would only want to get GPS-based location I can clean up permissions a bit.

EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/LocationFinder.java
```diff
@@ -0,0 +1,161 @@
...
+import android.location.LocationManager;
+import android.os.Bundle;
+import android.os.IBinder;
+import android.provider.Settings;
+import android.util.Log;
+
+import androidx.appcompat.app.AlertDialog;
+import androidx.core.app.ActivityCompat;
+
+public class LocationFinder extends Service implements LocationListener {
```
rightparen: Why does this extend Service?

EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/LocationFinder.java
```diff
@@ -0,0 +1,161 @@
...
+
+    @Override
+    public IBinder onBind(Intent intent) {
+        return null;
+    }
+    public Location getLocation() {
+        try {
+            locationManager = (LocationManager) context
+                    .getSystemService(LOCATION_SERVICE);
+// getting GPS status
```
rightparen: No need for self-evident comments.

EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/LocationFinder.java
```diff
@@ -0,0 +1,161 @@
...
+                    // Log.e(“Network”, “Network”);
+                    if (locationManager != null) {
+                        location = locationManager
+                                .getLastKnownLocation(LocationManager.NETWORK_PROVIDER);
+                        if (location != null) {
+                            latitude = location.getLatitude();
+                            longitude = location.getLongitude();
+                        }
+                    }
+                } else
```
rightparen: Are there cases where network-based-location is enabled but no network is in range so we have to fall back to GPS?

EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/LocationFinder.java
```diff
@@ -0,0 +1,161 @@
...
+        return longitude;
+    }
+    public boolean canGetLocation() {
+        return this.canGetLocation;
+    }
+
+    public void showSettingsAlert() {
+        AlertDialog.Builder alertDialog = new AlertDialog.Builder(context);
+        alertDialog.setTitle("GPS settings");
+        alertDialog.setMessage("GPS is not enabled. Do you want to go to settings menu?");
```
rightparen: "GPS is not enabled.  Click 'Settings' to enable."

EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/PhotoUploader.java
```diff
@@ -130,15 +130,19 @@ private void logException(String func, String location, Throwable err) {
...
-    try {
-      Path inputPath = filePath.toPath();
-      Path outputPath = Paths.get(uploadedDirPrefix + photoId + ".jpeg");
-      Files.move(inputPath, outputPath);
-    } catch (IOException ex) {
+
+      File sourceFile = new File(filePath.getAbsolutePath());
+      File destinationFile = new File(uploadedDirPrefix + photoId + ".jpeg");
+
+    if(sourceFile.renameTo(destinationFile)) {
```
rightparen: Interesting.  Were you seeing failures with `Files.move`?

cc @Ninjestic

goldjay: Yes, Files.move was added in API 26

EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/LocationFinder.java
goldjay: When I started this ticket, I intended it to act more like a service: stay on and update GPS location to use LastKnownLocation. At this point it's not acting much like a service so I can remove it.

EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/LocationFinder.java
goldjay: That's a good point. I'll check the internet connectivity as well as permissions.

rightparen: I'm not sure that's quite right.  You might have internet connectivity through cell data.

Would it make sense just to ask for both at once and then cancel once we have a decent radius from one of them?

goldjay: Yes! I pushed a changed to request updates from both and, in the case that network_provider doesn't do the job, location will come from GPS.

EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/LocationFinder.java
```diff
@@ -0,0 +1,149 @@
...
+                        LocationManager.NETWORK_PROVIDER,
+                        MIN_TIME_BW_UPDATES,
+                        MIN_DISTANCE_CHANGE_FOR_UPDATES, this);
+
+                locationManager.requestLocationUpdates(
+                        LocationManager.GPS_PROVIDER,
+                        MIN_TIME_BW_UPDATES,
+                        MIN_DISTANCE_CHANGE_FOR_UPDATES, this);
+
+                if(locationManager != null) {
```
audererob: I don't think this line is needed.

EbPhotoStoreNative/app/src/main/java/EbPhotoStoreNative/LocationFinder.java
```diff
@@ -0,0 +1,149 @@
...
+
+import static android.content.Context.LOCATION_SERVICE;
+
+public class LocationFinder implements LocationListener {
+    Context context;
+    boolean isGPSEnabled = false;
+    boolean isNetworkEnabled = false;
+    boolean canGetLocation = false;
+    Location location = null;
+    double latitude;
```
audererob: Should these be initialized to something?

goldjay: @audererob Do you mean latitude and longitude?

audererob: Yes

PR #90, opened 2019-08-20
Remove Unnecessary Files From flustudy_us

goldjay:
[Pull #76 ](https://github.com/AudereNow/audere/pull/76/files)

* Accidentally committed some files in flustudy_us in pull #76 
* Removed the files in question

I wanted to check that iOS and Android still worked after removing the files so I had to install the older version of cocoapods (1.5.3) and pod install on iOS. For android I had to change the version of google-play-services to match the version used at runtime.


rightparen:
@tpaik18 I created a list with this directory to track things we want to remove next time we do a rewrite.  150MB is a pain but I don't think it is worth rewriting for.

tpaik18:
@rightparen Okay, thanks.

PR #92, opened 2019-08-20
FEV-1117 Disable auto-advance from RDT screen on low memory warning, add event

audererob:
[FEV-1117](https://auderenow.atlassian.net/browse/FEV-1117)
We're seeing a decent number of instances of memory warnings causing RDT capture to advance to the manual page - this was put in to avoid the app crashing due to out-of-memory conditions. Since then, RDT capture has improved on its memory usage and this code isn't thought to be required. i.e. The app can perform successfully RDT capture under lower memory conditions. We'll disable that auto-advancement and add an event when the low memory event is received - we can then see how many of those cases made it past RDT capture to confirm that hypothesis.


FluStudy_au/src/ui/components/flu/RDTReader.tsx
```diff
@@ -81,7 +85,8 @@ interface FeedbackInstructionRequest {
 }
 
 const DEBUG_RDT_READER_UX = process.env.DEBUG_RDT_READER_UX === "true";
-const ALLOW_ICON_FEEDBACK = false; // Experimental for now; we'll revisit whether we should add icons back in
+const ALLOW_ICON_FEEDBACK = false; // Show iconic feedback during capture
+const ADVANCE_ON_MEMORY_WARNING = false; // Navigate to fallback page if memory warning occurs
```
rightparen: Should this be in firebase config?

smeds1: A Firebase config would be a good idea. Then we can revert without doing a new build if users are crashing.

PR #93, opened 2019-08-20
Commit WhatsApp data into Firebase

auderephilip:
This introduces WhatsApp EVD data into Firebase using the same schema as the RN app.  We fake some things (e.g. GPS location), but can do future commits that update what we can from user data.

image


EbPhotoSite/functions/src/index.ts
```diff
@@ -257,10 +287,178 @@ async function onNewCHW(agent: WebhookAgent) {
...
+
+  const docId = uuid();
+  const encounter = await createEncounterDoc(docId, chwMobile, photoId);
+
+  if (encounter) {
+    await admin
+      .firestore()
+      .collection("encounters")
+      .doc(docId)
+      .set(encounter, { merge: true });
```
billy-audere: I'm not sure if I'm missing a scenario where we'd actively want to merge encounter documents but I might prefer to overwrite.

EbPhotoSite/functions/src/index.ts
```diff
@@ -257,10 +287,178 @@ async function onNewCHW(agent: WebhookAgent) {
...
+  const localFile = await downloadUri(photoUri);
+
+  await admin
+    .storage()
+    .bucket()
+    .upload(localFile, {
+      destination: `photos/${photoId}.jpg`,
+      contentType: "image/jpeg",
+    });
+  fs.unlinkSync(localFile);
```
billy-audere: Might consider putting this in a finally block as the documentation around cleaning up temp storage from functions is rather ominous.

PR #95, opened 2019-08-20
Moving some classes over to a shareable backend library

billy-audere:
This would be in preparation for moving data pipelines out of FluApi.  The hope is to be able to move things out incrementally and preserve a shared data model across different application components for the time being.

Creates a backend-lib module which only contains a few core utils and data models but can be extended as needed.  I wanted to move SplitSql and the associated scaffolding for defining Sequelize models out of FluApi but preserve most of the existing data access code (several queries relying on Sequelize, a few earlier generations of data models).  As a result of some buggy behavior and our linking between local dependencies this caused an issue with Sequelize when loaded into the execution context multiple times.  There are SQL operators encapsulated as Symbols in the Sequelize module and using them across library boundaries breaks equality between Symbol operators, which breaks Op statements and similarly literals.

The approach I advocate here is using yarn workspaces to hoist the Sequelize module and flatten dependencies.  The workspace here is managed via Lerna and can be initialized by running `npx lerna bootstrap` from the root directory - which creates a shared node_modules directory at the root where the individual packages can recurse up to and resolve dependencies.


rightparen:
@billy-audere `react` is picky about having things installed in higher directories.  This diff ends up breaking EbPhotoSite's `yarn test`.  (I'll enter an issue to add that to CircleCI).

Might it make sense to create a subdirectory for the shared components with lerna there, have libraries subdirectories there that are actually shared, and then also put the bulk of the FluApi code there as well while we factor out the rest of the sequelize parts?

billy-audere:
@rightparen I'd be fine creating a subdirectory for shared components but don't know if that would be scalable for different application groupings that could arise.  It could also lead to an unintuitive layout for the repo.

I'm not in any hurry to get this merged, it has no real purpose until a larger set of infrastructure changes are ready.  I'll propose two thoughts I'd want to pursue:

- Use `nohoist` configuration in the workspace to prevent the libraries from being hoisted that are part of the preflight checks.  Fortunately it does seem like a [limited set](https://github.com/facebook/create-react-app/blob/f5b0aac4116fe58694db7c8996c4fd7da4821b9a/packages/react-scripts/scripts/utils/verifyPackageTree.js#L24-L30).
- See if Yarn Plug'n'Play can be used to circumvent the need for hoisting.

I'd appreciate any feedback - if you have strong feelings about either approach let me know.

rightparen:
Interesting.  No strong feelings, but yarn Plug'n'Play seems like it would be nice as long as it works.  E.g. it seems there is some concern over `tsc`'s ability to resolve things (https://yarnpkg.com/en/docs/pnp/getting-started).

PR #98, opened 2019-08-21
FEV-558 Adding license headers

Jaystar2323:
Takes in wildcards from the command line and can edit all those files recursively


scripts/update_copyright.py
```diff
@@ -39,19 +47,48 @@ def editFile(path):
         header_found = True
         break
     if not header_found:
-        print("Create new header")
+        if(len(text) == 0):
```
auderephilip: [tiny] Recommend this to happen at line 33 instead.  This code as-is is totally correct... but I think when reading the code (or thinking about it), it feels cleaner to have the logic be, "Read the whole file;  if it's empty, don't bother with anything else;  if it's not empty, look for a copyright..."

scripts/update_copyright.py
```diff
@@ -39,19 +47,48 @@ def editFile(path):
...
+            return
         if(text[0].startswith("#!")):
-            text.insert(1, header_text[0])
+            text.insert(1, createHeader(path))
```
auderephilip: Great work cleaning this up into a `createHeader` function.

scripts/update_copyright.py
```diff
@@ -39,19 +47,48 @@ def editFile(path):
...
-        #editFile(filepath)
-        #print(filepath)
-editFile(test_path)
+def createHeader(file):
+    header = ""
+    comment_type = "//"
+    is_block = False
+    if(file.endswith(".py") or file.endswith(".sh")):
+        comment_type = "#"
+    if(file.endswith(".css")):
```
auderephilip: [tiny] Could we eliminate the outer parens to be consistent with line 71?  (Also, are outer parens necessary on line 65 in python?)

scripts/update_copyright.py
```diff
@@ -39,19 +47,48 @@ def editFile(path):
...
+        header = "/* \n"
+    if file.endswith(".html"):
+        comment_type = "-->"
+        is_block = True
+        header = "<!-- \n"
+    for line in header_text:
+        header += (comment_type if not is_block else "") + line + "\n"
+    if is_block:
+        header += comment_type + "\n"
+    return header
```
auderephilip: This implementation is very good.  Great job on it.

scripts/update_copyright.py
```diff
@@ -39,19 +47,48 @@ def editFile(path):
...
+
+if(not(recursive is None)):
```
auderephilip: Recommend `if recursive is not None:` for a more typical pythonic way to do this (e.g. reads more like English).

scripts/update_copyright.py
```diff
@@ -39,19 +47,48 @@ def editFile(path):
...
+    for (subdir, dirs, files) in os.walk(recursive[0]):
+        for pattern in wildcard_pattern:
+            for filename in fnmatch.filter(files, pattern):
+                filepath = subdir + os.sep + filename
+                print(filepath)
+                editFile(filepath)
+else:
+    for i in range(len(sys.argv)-1):
```
auderephilip: Recommend extracting everything from this line and below into a function.  Something like `processOneDirectory`.  Then the logic starting line 81 would look something like:

```
if recursive is None:
  processOneDirectory(currentDir)
else:
  for (subdir...) in os.walk(etc):
    processOneDirectory(subdir)
```

Then in `processOneDirectory`, you'd do something like you already do below, except accounting for where to put `subdir` into glob's path.  And maybe using `args.inputfile` instead of `sys.argv` everywhere.

scripts/update_copyright.py
```diff
@@ -39,19 +47,48 @@ def editFile(path):
...
+    for line in header_text:
+        header += (comment_type if not is_block else "") + line + "\n"
+    if is_block:
+        header += comment_type + "\n"
+    return header
+
+if(not(recursive is None)):
+    for (subdir, dirs, files) in os.walk(recursive[0]):
+        for pattern in wildcard_pattern:
+            for filename in fnmatch.filter(files, pattern):
```
auderephilip: (I suspect by following the comment below, fnmatch might not longer be needed, and you could just use glob.  I think.  But I'm not positive.)

scripts/update_copyright.py
```diff
@@ -39,19 +47,48 @@ def editFile(path):
...
+                filepath = subdir + os.sep + filename
```
auderephilip: You'll typically want to do `os.path.join(subdir, filename)` to accomplish this.  That function not only understands os.sep, but also is smart enough to check against accidentally double-separating the final path (e.g. if subdir already has a slash at the end, which can happen in other types of cases, e.g. where the user passes in a path).

scripts/update_copyright.py
```diff
@@ -39,19 +47,48 @@ def editFile(path):
...
+            return
         if(text[0].startswith("#!")):
-            text.insert(1, header_text[0])
+            text.insert(1, createHeader(path))
         else:
-            text.insert(0, header_text[0])
+            text.insert(0, createHeader(path))
 
     f = open("test.txt", "w")
     f.writelines(text)
```
auderephilip: Recommend potentially printing out something like "Edited {path}" whenever you actually edit a file, so the person running it has a sense of what you've modified.

scripts/update_copyright.py
```diff
@@ -39,19 +47,48 @@ def editFile(path):
...
+        if(len(text) == 0):
+            return
         if(text[0].startswith("#!")):
-            text.insert(1, header_text[0])
+            text.insert(1, createHeader(path))
         else:
-            text.insert(0, header_text[0])
+            text.insert(0, createHeader(path))
 
     f = open("test.txt", "w")
```
auderephilip: Later, when ready (doesn't have to be this PR), we should turn this into "path" so that you make the edit into the actual file.

scripts/update_copyright.py
```diff
@@ -3,15 +3,22 @@
...
 
 now = datetime.datetime.now()
 header_text = [" Copyright (c) " + str(now.year) + " by Audere", "", 
 " Use of this source code is governed by an MIT-style license that", 
 " can be found in the LICENSE file distributed with this file."]
 
 parser = argparse.ArgumentParser()
-parser.add_argument("inputfile")
-test_path = parser.parse_args().inputfile
+parser.add_argument("inputfile", nargs='+')
```
auderephilip: [tiny] Recommend naming this `inputfilespec` instead, since it can be either a name or a filespec (e.g. with wildcard).

scripts/update_copyright.py
```diff
@@ -39,19 +48,45 @@ def editFile(path):
...
+    for pattern in wildcard_pattern:
+        for filename in glob(os.path.join(path, pattern)):
+            if not os.path.isdir(filename):
+                editFile(filename)
+
+if recursive is not None:
+    for (subdir, dirs, files) in os.walk(recursive[0]):
+        processDirectory(subdir)
+else:
+    processDirectory(".")
```
auderephilip: @Jaystar2323 : is "." interpreted as "the directory of the script" or "the directory that is the user's current directory?"

scripts/update_copyright.py
```diff
@@ -3,15 +3,21 @@
 import sys
 import re
 import datetime
+from glob import glob
+import fnmatch
```
auderephilip: I think this can be removed now...

scripts/update_copyright.py
Jaystar2323: I tested it, and I'm pretty sure it is the user's current directory

PR #100, opened 2019-08-21
Log detected test strip boundary points

balderdash:
The RDT reader reports the corners of where (it thinks) the test strip is. Log that to help ubicomp debug cases where it locates the test strip incorrectly.


FluStudy_au/android/app/src/main/java/host/exp/exponent/RDTReader.java
```diff
@@ -92,6 +97,16 @@ public void onRDTCameraReady(boolean supportsTorchMode) {
             event.putString("img", ImageUtil.matToBase64(captureResult.resultMat));
             event.putString("resultWindowImg", ImageUtil.matToBase64(interpretationResult.resultMat));
         }
+        if (captureResult.boundary != null) {
```
audererob: Is there a change to ImageProcessor.java that isn't here?

balderdash: Nope, boundary was already being included in captureResult, I just wasn't passing it along.

PR #101, opened 2019-08-22
Chills build setup

tpaik18:
Set Chills version to 1.0 and reset build number to 1. 
Fix config.yml where we were still putting access key A and B in .env for Cough. Now we will only do the .env creation for Sniffles or Fever. 
Add build_deploy_chills and build_deploy_chills_android targets, set up metadata for Play Store listing.


tpaik18:
Documentation of RE work needed along with this PR to get the new release set up on both stores: 
https://github.com/AudereNow/learn/wiki/How-to-set-up-a-new-release-on-App-Store,-Play-Store

rightparen:
@tpaik18 we will likely still need the A and B keys in Chills for address verification and kit pickup.

PR #102, opened 2019-08-22
Automation decline consent

smeds1:
The purpose of this PR is to restructure automation to allow for different paths through the app screens. To demonstrate the concept, I added an option to the test script that allows the automation to decline consent, return to the consent page, and then resume the journey through the app.

* Added "onClick" information to each continue button
* Restructured the script to use "onClick" information to determine next screen instead of going through the array of screens in order
* Added logic to allow denial of consent, return to consent page, and continuance with the app
* Added a pause at beginning of the app to prevent occasional crashes when the animation is still finishing
* Android apk path seems to have changed - adjusted accordingly


smeds1:
I also put in code to let automation enter invalid barcodes. If there are four invalid barcodes, automation tries each, reaches the "contact your clinic" screen, and then returns to the barcode entry screen to enter a valid code. If there are fewer than four invalid barcodes, automation tries each before entering the valid code.

PR #136, opened 2019-08-29
Copy Change for Negative Test Results

goldjay:
[FEV-1177](https://auderenow.atlassian.net/browse/FEV-1177)

* Added logic for if the user has a symptom that started 4 days ago
* Conditional UI in TestResult and TestResultRDT

Screen Shot 2019-08-29 at 10 21 02 AM


audererob:
I don't think we want any changes to the results screen for RDT interpretation - or @ShawnaCooper do we?

tpaik18:
My understanding @audererob and @ShawnaCooper is that we want the change to happen on the negative result screen for all paths (from blue/pink line answer, from high contrast answer, and from auto-interpretation). However, we do not want to change the positive result screen.

FluStudy_au/ios/Podfile.lock
```diff
@@ -885,4 +885,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 79bf68b93b7ee951ac47ac6648f9c06fc8b6bc4b
 
-COCOAPODS: 1.7.2
```
tpaik18: @goldjay The cocoapods version specified in the Gemfile is 1.7.2. If it's a good time for all of us to upgrade to cocoapods 1.7.5, then please update Gemfile and Gemfile.lock as well. Otherwise other engineers will keep showing Podfile.lock as changed on their machines.

FluStudy_au/src/i18n/locales/en.json
```diff
@@ -28,15 +28,20 @@
...
+      "negativeExplanation": "Sometimes people who do have the flu can receive a negative test result because:\n",
+      "negativeExplanationBullet": "The swab did not collect enough virus to indicate a positive result.",
+      "negativeExplanationBulletAll4Day": "Too much time passed between when you first felt sick and took the test.",
```
tpaik18: All4Day and OneDay seem to be counterintuitive labels. Maybe change All4Day to LessThan4Days, and OneDay to 4Days?

FluStudy_au/src/i18n/locales/en.json
```diff
@@ -28,15 +28,20 @@
     },
     "testResult": {
       "desc": "Your test strip indicates you may be:",
-      "positive": "**POSITIVE** for **INFLUENZA**",
-      "negative": "**NEGATIVE** for **INFLUENZA**",
+      "descNegative": "Your test strip indicates you **may be **",
+      "positive": "**positive for influenza**",
+      "negative": "**negative for influenza**",
```
tpaik18: Adding a . at the end of the sentence for the negative case would be nice.

FluStudy_au/src/i18n/locales/en.json
```diff
@@ -28,15 +28,20 @@
     },
     "testResult": {
       "desc": "Your test strip indicates you may be:",
-      "positive": "**POSITIVE** for **INFLUENZA**",
-      "negative": "**NEGATIVE** for **INFLUENZA**",
+      "descNegative": "Your test strip indicates you **may be **",
+      "positive": "**positive for influenza**",
```
tpaik18: I think we don't want to change the Positive UI at all.

ShawnaCooper:
Correct - we aren't changing the Positive results page - but now we have
two possible pages for the negative case. Please ping me via Slack if there
is confusion on the Jira ticket.

Thanks!
-Shawna

FluStudy_au/ios/Podfile.lock
goldjay: @tpaik18 I don't think there would be any harm in upgrading to 1.7.5, but it wasn't my intention to upgrade. I'll change my Podfile.lock back to 1.7.2.

FluStudy_au/src/i18n/locales/en.json
goldjay: I changed it to "AllUnder4Days" and "One4Days". LMK if that isn't clear.

PR #139, opened 2019-08-29
FEV-1074: Rash website localization

Ninjestic:


EbPhotoSite/src/PatientDetailPage.tsx
```diff
@@ -253,28 +271,33 @@ interface PatientInfoPaneProps {
...
   private renderDiagnosisBubble(diagnosis: Diagnosis) {
+    const { t } = this.props;
     const { value: evdPositive, diagnoser, timestamp } = diagnosis;
     const className = evdPositive ? "Positive" : "Negative";
-    const message = evdPositive
-      ? "*Likely POSITIVE for Ebola"
-      : "*Likely NEGATIVE for Ebola";
+    const message = evdPositive ? t("likelyPositive") : t("likelyNegative");
     return (
+      // TODOZ: figure out how to implement <strong> tag with tFunction
```
goldjay: We've got something in flustudy_au in the custom <Text /> component that bolds based on double asterisks like markdown. Maybe we could have something similar on the website.

EbPhotoSite/src/PatientDetailPage.tsx
```diff
@@ -405,11 +434,11 @@ class TriagePane extends React.Component<TriageProps, TriageState> {
         : triage && triage.notes;
     return (
       <div className="TriagePane">
-        <h3>Does the below image indicate Ebola positivity?</h3>
+        <h3>{t("triagePane:EVDPositivity")}</h3>
         <div className="EditDetail">
           <input
             type="button"
-            value="YES"
+            value={t("common:button:YES")!}
```
goldjay: Do you need the "!" ?

EbPhotoSite/src/PatientDetailPage.tsx
Ninjestic: The FluStudy_au Text.tsx is homebrewed and confusing and I'm afraid of even stating to implement it (I actually forgot about that comment and ended up putting in a different, potentially fine solution instead, let me know if figuring out text markdowns is important, and whether I should look into it or let someone else).

EbPhotoSite/src/PatientDetailPage.tsx
Ninjestic: It was throwing an error earlier due to it potentially being null, the "!" tells the compiler to not worry about it being null. Removing it now doesn't seem to cause problems though so I'll look into whether they seem safe to remove

EbPhotoSite/src/Chat.tsx
```diff
@@ -109,15 +113,17 @@ export class Chat extends React.Component<ChatProps, ChatState> {
 
         await api.pushNotification(
           doc.token,
-          `Message from ${currentUser.displayName}`,
+          t("messageFrom", { name: currentUser.displayName }),
+
           input,
           details,
-          "chw_chat"
+          t("messageGroup")
```
tpaik18: "chw_chat" looks like an internal string and not user-facing so probably shouldn't be localized. @balderdash ?

EbPhotoSite/src/PatientDetailPage.tsx
```diff
@@ -192,15 +194,16 @@ class PatientDetailPageAssumeRouter extends React.Component<
...
         await api.pushNotification(
           doc.token,
-          "Updated Ebola diagnosis",
-          "A patient's test result interpretation is available",
+          t("updatedDiagnosis"),
+          t("resultAvailable"),
           details,
-          "triage_evd"
+          t("messageGroup")
         );
```
tpaik18: "triage_evd" looks like an internal string and not user-facing so probably shouldn't be localized. @balderdash ?

EbPhotoSite/src/PatientDetailPage.tsx
```diff
@@ -592,7 +626,8 @@ class PhotoPane extends React.Component<PatientInfoPaneProps, PhotoPaneState> {
         {rdtPhotos.length > 1 && (
           <details>
             <summary>
-              Show previous {rdtPhotos.length > 2 ? "photos" : "photo"}
+              {t("showPrevious")}
+              {rdtPhotos.length > 2 ? "photos" : "photo"}
             </summary>
```
tpaik18: Looks like `photo` and `photos` is still hard-coded here in the string.

EbPhotoSite/src/Chat.tsx
balderdash: Yup, it should be left alone (or moved to a constant declared elsewhere, but that's out of scope for this PR).

EbPhotoSite/src/PatientDetailPage.tsx
balderdash: Yup, same as above.

EbPhotoSite/src/PatientDetailPage.tsx
tpaik18: @Ninjestic You could check out https://www.i18next.com/translation-function/plurals for best practices on plurals in i18n.

EbPhotoSite/src/i18n/index.ts
```diff
@@ -0,0 +1,32 @@
...
+// can be found in the LICENSE file distributed with this file.
+
+import * as Localization from "expo-localization";
+import i18n from "i18next";
+import enStrings from "./locales/en.json";
+
+const languageDetector = {
+  type: "languageDetector",
+  async: false,
+  detect: () => Localization.locale,
```
balderdash: You should be able to use `window.navigator.language` instead of `Localization.locale` here, in which case we won't need to add `expo-localization` to the project.

PR #140, opened 2019-08-29
Add form for setting "expert read" data on cough photos

balderdash:
Show a form alongside cough photos for Sam (or whoever else has permission) to enter their interpretation of a test strip. This data will be used by us and/or Ubicomp to assess the accuracy of automatic interpretation.

image

### DIFF
--- a/FluApi/src/endpoints/webPortal/rdtPhotos.ts
+++ b/FluApi/src/endpoints/webPortal/rdtPhotos.ts
@@ -7,6 +7,7 @@ import { Op, cast, json, where, col, fn } from "sequelize";
 import querystring from "querystring";
 import { SplitSql } from "../../util/sql";
 import { CoughModels, defineCoughModels } from "../../models/db/cough";
+import { AuthManager, Permissions } from "./auth";
 
 const LABELS = {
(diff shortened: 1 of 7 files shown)


FluApi/src/endpoints/webPortal/rdtPhotos.ts
```diff
@@ -138,6 +154,39 @@ export class RDTPhotos {
...
+      csrf: req.csrfToken(),
+      canInterpret,
+      interpretations,
+    });
+  };
+
+  public setExpertRead = async (req, res) => {
+    const { surveyId, interpretation } = req.body;
+    const interpreterId = req.user.id;
+    await this.models.expertRead.upsert({
```
rightparen: What does the `upsert` key off of?  I'd expect either `insert` here and allow multiple, or to have a uniqueness constraint on surveyId and interpreterId in sql.

balderdash: `upsert` uses the uniqueness constraint that's already on `surveyId`. I'm not saving an interpretation per-interpreter, I'm saving the interpreter id just so we have a record of who last updated it.

PR #148, opened 2019-09-03
EbPhotoSite: French strings courtesy of Google Translate

tpaik18:
Generated a fr.json for EbPhotoSite in order to test https://auderenow.atlassian.net/browse/FEV-1074

image


tpaik18:
Okay, the French site doesn't look like it has any glaring errors now based on my mediocre French. @auderephilip not sure if merging this in might have any possible demo value?

tpaik18:
Great work done by @Ninjestic to pull all the strings out. 
@auderephilip Merged. To see the site in French, change Chrome settings > Advanced > Language to add/set French the top language, then reload the page. If we don't demo this in the near future though, I would vote to take fr.json out because keeping strings in sync is such a pain.

auderephilip:
Ah, good to know.  Thanks for the extra context, @tpaik18.  Would it be workable to let the strings drift for now (if divergence happens) and for us to update them one more time if we get a big demo in the future?  It'd be ideal if we could minimize the number of times we need to keep strings in sync... (or if it's not easy/possible to batch up the updates to minimize that cost, LMK...)

tpaik18:
@auderephilip Okay yes that would work too.

PR #157, opened 2019-09-05
Trigger photo re-upload to S3 when it gets replaced

balderdash:


balderdash:
The originals are still in firebase, I think that's good enough for now.

PR #158, opened 2019-09-05
Importing Cough Qualtrics survey data

billy-audere:
Creates a new user & a bucket for uploading CSV files through the FTP server.  Lambda to import files is triggered when a new file is uploaded.  This uses mostly the same logic as ASPREN (PCR) data import and could probably be refactored a bit for reuse.

Database schema is a naive long list of mostly question-driven columns.  Sam and I discussed that this may be a sufficient initial approach since we don't anticipate questions changing substantially & there is a lot more engineering effort that would be needed to handle question versioning.

Finally the export into `cough_derived` is a * passthrough which should not be of much interest.


FluApi/src/services/cough/qualtricsImport.ts
```diff
@@ -0,0 +1,160 @@
...
+      "QID61" in survey &&
+      "QID45" in survey &&
+      "QID28" in survey &&
+      "QID62" in survey &&
+      "QID63" in survey
+    );
+  }
+
+  public async importFollowUpSurveys(): Promise<void> {
+    const file = await this.s3.getFollowUpSurveys();
```
balderdash: Since this always returns the most recent csv, is there any risk of accidentally skipping a file if two or more get uploaded in rapid succession?

FluApi/test/fever/s3Uploader.test.ts
```diff
@@ -13,6 +13,7 @@ describe("S3 uploader", () => {
   const s3Config: S3Config = {
     fluReportsBucket: "test_bucket",
     asprenReportsBucket: "not_test_bucket",
+    coughFollowUpBucket: "bucket_in_space",
```
balderdash: 🚀

FluApi/src/services/cough/qualtricsImport.ts
billy-audere: Yes, although I think that behavior may be desirable.  My assumption is that an export of Qualtrics data would likely be a full view of the survey responses so we wouldn't lose data if skipping a version.  That assumption needs to be validated before merging.

This will need to be updated if we need to process all files.  In that case we'd process all unprocessed files here.  At the route level we establish a SQL lock (there's a `sqlLock.runIfFree` wrapper) which should mean it is safe to process multiple files in a single call as there should ever only be one action in flight modifying the database.

I'll follow up.

billy-audere: @balderdash The reports are cumulative so this looks good.  I added record deletion so its a full sync and so I can easily import some dummy test records and overwrite them.

PR #159, opened 2019-09-06
FEV-1213: Restore auto-navigation path for PubHubSub

smeds1:
This PR restores the auto-navigation path for the PubHubSub package. 

Now that the TestResult screen requires answers to the "how long ago did your symptoms start" questions, the auto-navigation crashes on that screen. I cheated a bit and re-routed the navigation to skip that screen for now.


auderephilip:
thanks for fixing this, @smeds1 .  LMK if you or @tpaik18 find this useful over time.  when i originally  wrote it, i thought i'd use it more... but i still manually click through every screen.  the code is a bit hacky and fragile, and so maintaining it might not be worth it unless you or other members of the team find it useful.

smeds1:
@auderephilip I don't think anyone is using this package at the moment, but we thought it might be useful for automation. Right now we are struggling with doing any serious automated testing of the RDT reader since it takes several minutes for Appium to arrive at that screen. I'm going to try doing the fast-forward with PubHubSub instead (since it's so much quicker) and see if we can get some more sophisticated RDT reader testing up and running.

PR #160, opened 2019-09-06
FEV-1208 If RDT captured but is still processing at 30 seconds, the app should not switch to manual photo

audererob:
If we receive the onRDTInterpreting event, we'll ensure that the timeout handling doesn't occur.


tpaik18:
@audererob Did you mean to add FluStudy_au/buildInfo.json.new with this PR?

audererob:
No - that file has showed up in my past few PRs and I've typically caught it. Any idea what it is? I can remove it.

tpaik18:
@audererob `yarn start` calls `../scripts/generate-build-info` which calls `../scripts/extend-build-info` which does that `mv` from buildInfo.json.new to buildInfo.json. Do you see any console error when running `yarn start` about the `mv` failing for some reason?

tpaik18:
Yes, I was thinking when I saw Ram's https://github.com/AudereNow/audere/pull/118 that people might not have `jq` installed.

audererob:
11d9b95a6f9acc1ddc8895f025247faced291b2d should fix this.

PR #161, opened 2019-09-06
Automation rdt stats

smeds1:
The purpose of this PR is to report whether or not RDT capture succeeds. In the case of a successful capture, the rdtInfo should be displayed in the console. I also added a pause and a longer initial timeout to combat some gremlins. 

image


smeds1:
image

PR #166, opened 2019-09-09
FEV-1206 Timers occasionally freeze at 0:01

audererob:
[FEV-1206](https://auderenow.atlassian.net/browse/FEV-1206)

This is fix represents a hypothesis as to the root cause of the issue (i.e. no definitive cause was found, so this is a guess). The hypothesis is that the issue is in the _onTimer method. Formerly, it was calling _getRemainingLabel followed by _getRemainingMs. Each of these calls uses the current clock value as a computation - thus it is possible (and likely?) that the clock values will be different across those calls. If the clock value happens to change from > 1 second to <= 1 second between those calls, then the timer will be stopped but remainingLabel will still be 0:01. It won't get updated and won't trigger a re-render which would otherwise output the continue button. To fix this, the remaining ms is computed first and then the remaining label is updated. That original remaining ms is then checked and the timer cancelled if needed.

A glitch in this logic is that before any fix mentioned here, setState is getting called in _onTimer which I believe *should* cause a re-render after the method completes, causing the continue button to be output. I'm not sure if it's possible for the timer to compute two consecutive equal values for remainingLabel, or if setState can be applied right away (i.e. before _onTimer exits) to result in no render occurring, thus the uncertain nature of this PR.


FluStudy_au/src/ui/components/Timer.tsx
```diff
@@ -133,8 +133,9 @@ class Timer extends React.Component<Props & WithNamespaces> {
 
   _onTimer = () => {
     if (this.props.navigation.isFocused()) {
+      const isDone = !this._getRemainingMs();
       this.setState({ remainingLabel: this._getRemainingLabel() });
```
auderephilip: super-pedantic here, but it seems that it'd be ideal if _getRemainingLabel() and _getRemainingMs() were somehow atomic (e.g. _getRemainingLabel() returning { label: boo, ms: int}).  But not needed here;  just a random musing.

audererob: Agreed, or if _getRemainingLabel was instead _getRemainingLabelForTime(time: number), passing in the result of _getRemainingMs.

PR #167, opened 2019-09-10
FEV-1097 Sometimes clicking the "App Version" menu link makes the version info appear in the background but menu stays open

audererob:
[FEV-1097](https://auderenow.atlassian.net/browse/FEV-1097)
We were persisting some navState that wasn't unique across relaunch, and that was confusing the drawer code. If we remove those values at persistence time, then this issue is resolved.


FluStudy_au/src/ui/ConnectedRootContainer.tsx
```diff
@@ -87,9 +87,14 @@ interface Props {
 }
 
 const persistenceKey = "NavigationStateAus";
-const persistNavigationState = async (navState: NavigationState) => {
+const persistNavigationState = async (navState: any) => {
```
goldjay: Why did you change navState from NavigationState to any?

audererob: Correct - there's a DrawerNavigationState but it doesn't have those properties. Looking at the source for the drawer router, it didn't look like there was anything strongly typed. I'll take another look to be 100% certain.

PR #176, opened 2019-09-11
FEV1232: Moved tap/scroll/find functions into separate script

smeds1:
This PR moves the tap/scroll/find functions from appium.test.js to util/commands.js. Here are some things I'm wondering about: 

* Is commands.js a reasonable name, or is there something more descriptive that makes sense?
* Should I also move the functions like android_select (pick a choice from an Android dropdown) to util? Would those functions go in the commands.js file or a separate file?
* Are there other functions from appium.test.js that should be moved?


tpaik18:
Oooh nice. Another idea--have you considered just passing down deviceInfo since the parameter list is getting kinda long?

smeds1:
The new multi_tap function is maybe a little flakey on Android. I may revert to old functions if it becomes problematic.

smeds1:
I think I now have all the question-type logic in the controls file. I believe there is more clean-up that I can do on some of those screen-specific functions, but I'll do that in another PR so that this one doesn't get out of hand.

PR #180, opened 2019-09-11
FEV-1221 Set GIFTCARD_KEY in cough circle build

balderdash:


balderdash:
On @tpaik18's suggestion, just going to use the existing document upload key instead.

PR #182, opened 2019-09-12
Cough incentive 

goldjay:
[FEV-1186](https://auderenow.atlassian.net/browse/FEV-1186)

One big change in App.tsx:  The remote configs were sometimes not being loaded before the initial load of the application. This is only a problem because I had text that needed something from the remote config on the first page. Otherwise I would have two different messages from Chrome and AnimatedChrome. I haven't seen any problems with anything loaded in "_loadAssets" i.e. fonts or the store, but if anyone thinks this is too much of a change I can work around it.

State A
STATE A

State B
STATE B

State C
STATE C

State D
STATE D

State E: Offline
STATE E

State F: Invalid Barcode
STATE F

State G: API Error or Invalid Install ID
STATE G


smeds1:
@goldjay Do we have any way to tell if the user was offered a gift card? I see that you're logging when a user pressed the "claim your card" button, but we probably also want to know if that button was there to begin with.

goldjay:
@smeds1 Yes, if the gift cards available switch is on, its connected to the internet and there are no api or invalidBarcode errors the gift card button is shown and I could log something. Would you want it to log each time it is shown or just once to know it has been shown?

smeds1:
@goldjay Let's definitely log something instead of trying to trace through all the conditions afterwards. I don't think it matters if we log once or multiple times.

FluStudy_au/src/i18n/locales/en.json
```diff
@@ -143,7 +143,9 @@
...
...eplace professional medical advice. The interpretation of your result may differ from a medical test conducted in a clinical lab environment. In no circumstances should the results of the test kit be relied upon without independent consideration and confirmation by a qualified medical practitioner.\n\nAudere makes no representations or warranties with respect to any treatment, action, suitability or application of medication or preparation by any person whether in connection with this app, the test kit or not. In no circumstances will Audere be liable for any direct, indirect, consequential, special, exemplary or other damages arising therefrom.\n\n**How much time will my involvement in the project take?**\nIt is expected that self-testing, completing the questionnaire, and reading results should take about 20 minutes.\n\n**Are there any risks associated with participating in this project?**\nSome people may feel uncomfortable taking a swab test from their nose. Rarely this might cause a small nosebleed. Since the results of the self-test kit are experimental, this may make people anxious. If you are worried about flu or your health you should seek advice and care from your GP.\n",
```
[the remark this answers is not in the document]
goldjay: @tpaik18 It's the easiest way I've found. If I separate the conditional sentence into its own text component the styling gets complicated and adding a custom function to conditionally insert one sentence over the other would also be overly complicated.

FluStudy_au/ios/Podfile.lock
```diff
@@ -885,4 +885,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 79bf68b93b7ee951ac47ac6648f9c06fc8b6bc4b
 
-COCOAPODS: 1.7.2
```
tpaik18: @goldjay Would you mind reverting this to 1.7.2? It is tripping up automation per @smeds1 because the version doesn't match. Thanks.

goldjay: @tpaik18 My bad, it keeps on making me pod install when I want to run it from a fresh branch.

tpaik18: @goldjay To avoid the problem I believe you can use `bundle exec pod install` which will always use the version of cocoapods specified in the Gemfile.

goldjay: @tpaik18 Nice!

PR #183, opened 2019-09-12
FEV-1221 API to retrieve cough giftcard URLs

balderdash:
This adds a table for storing giftcard URLs, and an API endpoint for the client to hit that validates the installationId and returns a giftcard with the appropriate denomination. There's also a checkGiftcardAvailability call that looks for giftcards without allocating one to the provided barcode.

The plan is to put Prezzee's demo URLs into the staging giftcard table, and actual URLs go in prod. If you're running in demo mode, you'll just get an example.com URL.

Followup PRs will include:
- Validating barcodes
- Populating the giftcard table from CSVs
- Adding data to cough_derived so that Monique can monitor giftcard usage


FluApi/src/util/sqlLock.ts
```diff
@@ -35,9 +35,27 @@ export class SqlLock {
...
+        logger.info(
+          `Acquired SQL lock '${scope}' (${key.hex64}), running callback`
+        );
+        const ret = await won(...args);
+        logger.info(`Releasing SQL lock '${scope}' (${key.hex64})`);
+        // Lock should be released when transaction completes.
+        return ret;
+      });
+    };
+  }
```
rightparen: 👍

PR #192, opened 2019-09-14
Add Flows prototype app

auderephilip:
In working with potential vNext partners, this is a basic project that introduces a React app that supports Firebase login.  It dynamically reloads the web app when you make changes and save;  it's linked with a new Firebase project for hosting and for cloud functions.  It supports email and Google authentication.

After this initial commit, more serious functionality will come.

Firebase docs confirm that checking in the API key is safe — that the Firebase config is not considered a secret.


tpaik18:
I meant to choose "Comment" and not "Approve"

auderephilip:
[just closing on this based on in-person convos with @rightparen ]  My understanding is that the key isn't a meaningful secret when you have auth-gated reads & writes for everything (which should be the case here).  In the case of the incentives project we're building, reads & writes will all require auth, and all auth'd accounts will be manually created by us, so we should be ok.

PR #194, opened 2019-09-16
Conditional rendering for the menu based on gift card availability

goldjay:
[FEV-1218](https://auderenow.atlassian.net/browse/FEV-1218)

Since the menu was all contained in the configs, it was getting rendered before the app had fetched the remote configs. My solution was to change the menu setup into a series of functions that would get executed at runtime. I haven't seen any slowdown at startup, but let me know if it doesn't seem right.


FluStudy_au/src/resources/MenuConfig.ts
```diff
@@ -36,118 +37,135 @@ function menuScreen(
...
-        titleLabel: q,
-        bodyLabel: q + FAQ_ANSWER_SUFFIX,
-        namespace: key,
-        appEvent: AppEvents.FAQ_PRESSED,
-      },
-    };
+  const giftCardsAvailable = getRemoteConfig("giftCardsAvailable");
+  let FAQComponents: Component[] = [];
+  FAQS.forEach((q, index) => {
+    if ((!giftCardsAvailable && index < 12) || giftCardsAvailable) {
```
audererob: Instead of hard-coding 12, you could add an optional boolean property to each item named "requiresGiftCardsAvailable", adding it to items at index 12 and above.

goldjay: @audererob That's a good idea!

FluStudy_au/src/i18n/locales/en.json
```diff
@@ -115,7 +115,17 @@
...
     "appDelete": "**After I finish the tests, can I delete the app?**",
-    "appDelete++Answer": "We will send you a single notification on your device a few days after you complete the test with a link to complete a survey about your experience using the app. Once you have received that notification and completed the survey you may delete the app."
+    "appDeleteGiftCard": "**After I finish the tests, can I delete the app?**",
+    "appDelete++Answer": "We will send you a single notification on your device a few days after you complete the test with a link to complete a survey about your experience using the app. Once you have received that notification and completed the survey you may delete the app.",
+    "appDeleteGiftCard++Answer": "Please complete the follow-up survey and claim your gift card before you uninstall this app. We’ll send you a link to the survey 1-2 days after you complete the test.",
```
tpaik18: Is there a reason we changed this to "1-2 days"? If we don't like "a few days" as it was previously, can we say "a couple days" so it matches the "Follow-up Survey" screen copy?

FluStudy_au/src/util/remoteConfig.ts
```diff
@@ -27,7 +27,7 @@ interface RemoteConfig {
 const DEFAULT_CONFIGS: RemoteConfig = {
   advanceRDTCaptureOnMemoryWarning: false,
   giftCardsAvailable: false,
-  giftCardAmount: "999",
+  giftCardAmount: "15",
```
tpaik18: I believe we want this to be 20 currently.

goldjay: @tpaik18 Okay, I'll make the change.

FluStudy_au/App.tsx
```diff
@@ -52,6 +52,7 @@ export default class App extends React.Component {
     // parallelize those two.
     await loadAllRemoteConfigs();
```
audererob: Just curious, did you try moving this above this._loadAssets? Does it require other stuff to be set up?

FluStudy_au/src/i18n/locales/en.json
goldjay: @tpaik18 That sounds like a good idea for consistency. I don't recall why we changed it, but maybe @ShawnaCooper would know.

FluStudy_au/App.tsx
goldjay: @audererob I haven't tried moving it above this._loadAssets. From Michael's comment about possibly parallelizing it seems like they don't depend on each other.

goldjay: @audererob Tried it out and fonts and whatnot need to be loaded before the remoteConfigs.

FluStudy_au/src/i18n/locales/en.json
```diff
@@ -115,7 +115,17 @@
...
     "appDelete": "**After I finish the tests, can I delete the app?**",
-    "appDelete++Answer": "We will send you a single notification on your device a few days after you complete the test with a link to complete a survey about your experience using the app. Once you have received that notification and completed the survey you may delete the app."
+    "appDeleteGiftCard": "**After I finish the tests, can I delete the app?**",
+    "appDelete++Answer": "We will send you a single notification on your device a couple days after you complete the test with a link to complete a survey about your experience using the app. Once you have received that notification and completed the survey you may delete the app.",
+    "appDeleteGiftCard++Answer": "Please complete the follow-up survey and claim your gift card before you uninstall this app. We’ll send you a link to the survey a couple of days after you complete the test.",
```
tpaik18: Sorry one more nit-pick here. Can we change "We'll send you a link to the survey" to "The app will send you a link to the survey"? This makes it clearer why we're asking them to wait for that before uninstalling.

goldjay: @tpaik18 Sure, that makes sense. I'll make the change.

PR #197, opened 2019-09-17
FEV-1234 update_copyright.py should support .gitignore, .gradle, .h, .java, .tf 

tpaik18:
Add support for .gitignore, .gradle, .h, .java, .tf extensions in update_copyright.py. 
.gradle, .h, .java, and .tf can use the default // comment and .gitignore uses the # comment. 
update_copyright.py inserted its own copyright header after I ran it in the `/scripts` dir as a test.


scripts/update_copyright.py
```diff
@@ -1,3 +1,7 @@
+# Copyright (c) 2019 by Audere
+#
+# Use of this source code is governed by an MIT-style license that
+# can be found in the LICENSE file distributed with this file.
```
rightparen: So meta...

PR #200, opened 2019-09-17
FEV-1221 Check for card availability in demo mode

balderdash:
Even though we aren't returning a real giftcard url if the request was sent with a demo mode flag, we still want to check that there are giftcards available to facilitate testing.


FluApi/src/endpoints/coughGiftcardApi.ts
```diff
@@ -103,28 +103,27 @@ export class CoughGiftcardEndpoint {
...
     if (!valid) {
       return { failureReason: validationFailureReason };
     }
-    const { giftcard, isNew, failureReason } = await this.getCard(
-      request,
-      true
-    );
+    const { giftcard, isNew, failureReason } = request.isDemo
+      ? await this.getCard(request)
+      : await this.getAndAllocateCard(request);
```
rightparen: Or just `this.getCard(request, !request.isDemo)`?

FluApi/src/endpoints/coughGiftcardApi.ts
```diff
@@ -103,28 +103,27 @@ export class CoughGiftcardEndpoint {
...
+      if (request.isDemo) {
+        return {
+          giftcard: {
+            url: DEMO_GIFTCARD_URL,
+            denomination: request.denomination,
+            isDemo: true,
+            isNew,
+          },
+        };
+      }
```
rightparen: I'm a little concerned with the code complexity for this.

Maybe something for a later PR, but would it make sense to have a demo flag on gift card rows in the db, and to allocate/query for cards that match the demo flag of the request?  That would let us use code that was identical on both client and server, and the demo flag on the client would flow all the way to the db and select there.  We could exercise running out of cards, getting different cards for different requests, verify that you get the same card if you repeat the request, etc. in demo mode.

If we did go that way, we could populate the demo cards via sql or dbcli and shouldn't need to extend the CSV importer.

balderdash: I actually went down that route initially, and I now regret changing my mind. I'm going to go back to that.

PR #213, opened 2019-09-18
FEV-1263: added logging for gift card failures

smeds1:
The purpose of this PR is to provide logging for all failures received from the gift card API. I was able to confirm that the API error is logged correctly.


goldjay:
@smeds1 When checking giftCardAvailability, there is a chance the failureReason could be because we are out of giftcards, api error or an invalid install id. Would it be helpful to be more specific when logging the api error?

smeds1:
@goldjay I am logging the numerical code that goes along with failureReason, so we should be able to differentiate errors when looking at the logging. Not elegant, but I think something all-purpose is better while we are in development.  

logFirebaseEvent(AppHealthEvents.GIFTCARD_API_ERROR, {
        failureReason: response.failureReason,
      });

smeds1:
@tpaik18 @goldjay Are you both ok with this change? We are now only logging the gift card button click if there is a valid card to claim. Otherwise we are logging an error. I also changed the logging to work in both demo and non-demo mode.

goldjay:
@smeds1 Yeah, seems like a good fix!

tpaik18:
Thanks @smeds1. Could we also remove the `!isDemo &&` on line 81?

smeds1:
@tpaik18 Good catch.

PR #215, opened 2019-09-19
Gift Card API Error Handling

goldjay:
[1260](https://auderenow.atlassian.net/browse/FEV-1260)

* Try / Catch blocks for API calls
* Error triggers error state change a la state G in FEV-1186


smeds1:
@goldjay This looks like what I want, but I'll wait for Jenny or Rob to give the official approval.

PR #216, opened 2019-09-19
No Email Support Button When Offline

goldjay:
[FEV-1265](https://auderenow.atlassian.net/browse/FEV-1265)


tpaik18:
@goldjay Not positive, but it seems like if I'm offline when component is mounting, then my failure to check gift card availability will be considered API_ERROR. I think this should not be considered API_ERROR.

goldjay:
@tpaik18 That's a good point. I'll add a check for connectivity before logging the API_ERROR.

FluStudy_au/src/ui/components/GiftCard.tsx
```diff
@@ -70,10 +70,12 @@ class GiftCard extends Component<Props & WithNamespaces, State> {
         this.setState({ failureReason: response.failureReason! });
       }
     } catch (e) {
-      logFirebaseEvent(AppHealthEvents.GIFTCARD_API_ERROR, {
-        failureReason: GiftcardFailureReason.API_ERROR,
-      });
-      this.setState({ failureReason: GiftcardFailureReason.API_ERROR });
+      if (!isConnected) {
```
tpaik18: Shouldn't this be `if (isConnected)` ?

tpaik18:
Thanks @goldjay. I hope this also resolves the part of FEV-1265 where if I become connected then the screen still shows me "We are experiencing an issue"?

PR #223, opened 2019-09-20
FEV-447 Post slack notifications from deploy-api

balderdash:
Post basic "Deploying"/"Finished Deploying" messages using Slack's simple [Incoming Webhooks API](https://api.slack.com/incoming-webhooks).


scripts/util.sh
```diff
@@ -5,6 +5,13 @@
 
 set -euo pipefail
 
+REPO_ROOT_DIR="$(git rev-parse --show-toplevel)"
+if [ -e "$REPO_ROOT_DIR/scripts/.env" ]; then
+  set -o allexport
```
rightparen: TIL

scripts/util.sh
```diff
@@ -91,3 +98,12 @@ function jqraw() {
 function echo2() {
   echo 1>&2 "$*"
 }
+
+function notify-slack() {
+  if [ -z ${SLACK_WEBHOOK+x} ]; then
```
rightparen: This is a version of the syntax that is a bit more explicit and searchable:
```
if [[ -z "${SLACK_WEBHOOK:-}" ]]; then
```

PR #231, opened 2019-09-23
Offline to Online on Final Screen

goldjay:
[FEV-1271](https://auderenow.atlassian.net/browse/FEV-1271)

Checks connectivity before showing an error state on redeem button press since connectivity is a separate issue.


tpaik18:
Hoping someone else can review this? Can't tell whether this code change addresses my bug.

PR #234, opened 2019-09-23
SurveyLinkBlock not appearing consistantly

goldjay:
[FEV-1273](https://auderenow.atlassian.net/browse/FEV-1273)

* SurveyLinkBlock handles its own conditional rendering, but needs to "null render" once in order to set its appstate event listener so I removed SurveyLinkBlock from the conditional rendering block in GiftCard


FluStudy_au/src/ui/components/GiftCard.tsx
```diff
@@ -181,12 +181,8 @@ class GiftCard extends Component<Props & WithNamespaces, State> {
...
           <Text content={t(thankYouText)} />
           <Divider style={styles.divider} />
-          {!!completed48HoursAgo && (
-            <Fragment>
-              <SurveyLinkBlock />
-              <Divider style={styles.divider} />
-            </Fragment>
-          )}
+          <SurveyLinkBlock />
+          {!!completed48HoursAgo && <Divider style={styles.divider} />}
```
audererob: Couldn't you just move the appState handling to this component and remove it from SurveyLinkBlock? We'll need that to fix the other bug where gift card availability isn't checked on foregrounding. You otherwise have two places that compute the same state (though it appears there will be differing results between 48 hours and one millisecond through 49 hours due to the differing time units they check).

PR #238, opened 2019-09-24
Deleting rows from derived tables when there is no matching source row

billy-audere:
Without cascading deletes to the derived table this should be generating duplicate values as well as preserving events that may have been modified or removed.

In current usage, for Firebase only, this generally this doesn't have much of an impact since the event itself is a shell - most of the data is stored as event params or user properties which are separate database objects.  However, the impact could have been much more significant if the `SequelizeSourceTable` was used elsewhere.


billy-audere:
Specifically for Firebase there is a transition from a temporary file (intraday) that causes almost every event to be recreated in the source table.  This leads to duplicates.

Also of note while looking at this I noticed that `event_timestamp` changes quite frequently for Firebase events.

PR #241, opened 2019-09-24
Offline to Online GiftCards

goldjay:
[FEV-1271](https://auderenow.atlassian.net/browse/FEV-1271)

* Final screen was giving an API error based on a failed request for gift card availability that was given while there was no internet connection
* Added a check for connectivity before passing the API error to local state


FluStudy_au/src/ui/components/GiftCard.tsx
```diff
@@ -68,7 +68,7 @@ class GiftCard extends Component<Props & WithNamespaces, State> {
         isDemo
       );
 
-      if (response.hasOwnProperty("failureReason")) {
+      if (response.hasOwnProperty("failureReason") && isConnected) {
```
audererob: Would it make sense to wrap the try/catch in a if (isConnected) condition, and the same for redemption?

goldjay: Yes, that makes sense.

FluStudy_au/src/ui/components/GiftCard.tsx
```diff
@@ -60,26 +60,28 @@ class GiftCard extends Component<Props & WithNamespaces, State> {
...
-        });
-        this.setState({ failureReason: GiftcardFailureReason.API_ERROR });
+        if (response.hasOwnProperty("failureReason")) {
+          logFirebaseEvent(AppHealthEvents.GIFTCARD_API_ERROR, {
+            failureReason: response.failureReason,
+          });
+          this.setState({ failureReason: response.failureReason! });
+        }
+      } catch (e) {
+        if (isConnected) {
```
audererob: Great! You should be able to remove the other isConnected conditions within the two blocks.

PR #246, opened 2019-09-25
Set Gift Card Amount For Pre-Incentive Returning Users

goldjay:
[FEV-1276](https://auderenow.atlassian.net/browse/FEV-1276)

Added a call to setGiftCardAmount on returning to the app if giftCardAmount is undefined.


FluStudy_au/src/ui/components/GiftCard.tsx
```diff
@@ -107,7 +108,14 @@ class GiftCard extends Component<Props & WithNamespaces, State> {
     if (nextAppState === "active") {
       // Force a re-render in case remoteConfig values have changed;
       // e.g. giftCardsAvailable or skipSurveyNotification
-      this.forceUpdate();
+      if (
+        this.props.giftCardAmount === undefined ||
+        this.props.giftCardAmount === ""
+      ) {
+        this.props.dispatch(setGiftCardAmount());
+      } else {
```
tpaik18: Would this result in setting gift card amount even if giftCardsAvailable==false?

goldjay: Yes, that's a good point. I'll add a check for giftCardsAvailable.

tpaik18: Also, not sure but wondering if that whole block between lines 66 and 100 should be inside an `if giftCardsAvailable` block?

goldjay: I suppose so. If gift cards were not available, then it doesn't matter if we are out of gift cards.

tpaik18:
Thanks @goldjay, hoping someone else can review the overall approach. I'm not sure what the ramifications are of moving forceUpdate() into the `else` block or making setGiftCardAmount trigger a sync.

PR #247, opened 2019-09-25
FEV-973 Add sex and age to rash patient info

balderdash:
You probably want these fields to be available while you're working on https://auderenow.atlassian.net/browse/FEV-972


goldjay:
@balderdash Actually I've already got them in a [PR](https://github.com/AudereNow/audere/pull/225). Though "sex" is probably more accurate than "gender" for what we are asking.

PR #248, opened 2019-09-25
Split Medication Questions

goldjay:
[FEV-1294](https://auderenow.atlassian.net/browse/FEV-1294)

* Added antiviral question 
* Changed copy on antivirus question
* Changed ID for antibiotics question to MedicationAntibiotics
* Changed ID for researchByAnyResearchers to researchByAllResearchers since the question was changed previously and we wanted different keys
Medicine Questions


tpaik18:
This PR seems okay. At first I thought the en.json key needed to change, but Sam pointed out that the `id` is what gets saved in the JSON `responses` section.

tpaik18:
@balderdash @audererob Okay to land this one?

PR #249, opened 2019-09-25
FEV-1292 Adding offline storage/caching for specified remote config values

audererob:
[FEV-1292](https://auderenow.atlassian.net/browse/FEV-1292)

Adding the ability for remoteConfig values to be cached/persisted locally in the case the network is offline. Also changing the type of giftCardAmount to number, since that's what type it is in the firebase console remote config.


FluStudy_au/src/store/survey.ts
```diff
@@ -19,7 +19,7 @@ import i18n from "i18next";
 import { getRemoteConfig } from "../util/remoteConfig";
 import { onCSRUIDEstablished } from "../util/tracker";
 
-export const DEFAULT_GIFT_CARD_AMOUNT = "";
```
goldjay: I think we can remove DEFAULT_GIFT_CARD_AMOUNT or set it to the actual giftCardAmount in the default configs. This was left over from a time when I was initializing giftCardInfo to an object with empty values for each property in an effort to tell if it has been set or not.

FluStudy_au/src/util/remoteConfig.ts
```diff
@@ -52,6 +56,51 @@ async function loadConfig(): Promise<RemoteConfig> {
...
+
+        logFirebaseEvent(AppHealthEvents.REMOTE_CONFIG_ERROR, {
+          errorMessage,
+        });
+        crashlytics.log(errorMessage);
+      }
+    } else {
+      // We're not connected to a network which means we should use the most
+      // recently persisted values.
+      (await AsyncStorage.multiGet(OFFLINE_PROPERTIES)).map(pair => {
```
goldjay: NIT: Maybe use forEach or a normal for loop (fastest) since we are not putting anything back into the array, just iterating over it.

tpaik18:
Looks okay, but is there a reason why we only want to cache the gift-card-related values? Would it hurt to cache the other values as well? Seems reasonable that if I launch without internet, the app would use the remote config values that it last used when I was online.

audererob:
@tpaik18 I was being conservative impact-wise, but you have a good point that we should persist for everything.

PR #251, opened 2019-09-26
FEV-1297: Restore deprecated questions

smeds1:
* Added researchBySameResearchers and antibiotics to coughQuestionConfig.ts and dist/CoughQuestionConfig.js
* Updated some old researchBySameResearchers labels that I missed yesterday to be researchByAnyResearchers 
* Added MedicationAntivirals to SurveyQuestions so that the new question will appear in Metabase


lib/coughQuestionConfig.ts
```diff
@@ -58,6 +58,17 @@ export interface DropDownQuestion extends SurveyQuestion {
...
+    { key: "yes", primary: false, enabled: true },
+    { key: "no", primary: false, enabled: true },
+  ],
+  id: "ResearchBySameResearchers",
+  required: true,
+  title: "obsolete question - do not use",
+  type: SurveyQuestionType.ButtonGrid,
+};
+
+export const ConsentAnyResearchersConfig: SurveyQuestion = {
```
rightparen: `ScreenConfig.ts` still references `ConsentSameResearchersConfig` and does not reference this one.  Given the contents, it looks like you mean for this to be the one that is active though.

lib/coughQuestionConfig.ts
```diff
@@ -990,6 +1001,18 @@ export const InterferingConfig: SurveyQuestion = {
   type: SurveyQuestionType.ButtonGrid,
 };
 
+export const AntibioticsOldConfig: SurveyQuestion = {
```
rightparen: I like this idea.  Let's name the Config objects `Obsolete...Config` so that it's really obvious they shouldn't be used.

PR #254, opened 2019-09-26
FEV-1275 Gift card limit is not checked when app is foregrounded

audererob:
[FEV-1275](https://auderenow.atlassian.net/browse/FEV-1275)
The only time that the checkGiftCardAvailability API was being called was on component mount, leading to this issue. This change refactors the code into a common function that will be called on component mount and on app foregrounding.


FluStudy_au/src/ui/components/GiftCard.tsx
```diff
@@ -58,7 +58,31 @@ class GiftCard extends Component<Props & WithNamespaces, State> {
...
+      this._checkGiftCardAvailability();
+
+      // Force a re-render in case remoteConfig values have changed;
+      // e.g. giftCardsAvailable or skipSurveyNotification
+      if (
+        !!getRemoteConfig("giftCardsAvailable") &&
+        !this.props.giftCardAmount
+      ) {
+        this.props.dispatch(setGiftCardAmount());
+      } else {
```
tpaik18: In the off-chance that giftCardAmount is 0 and we have to call setGiftCardAmount(), wouldn't we want to do that before calling checkGiftCardAvailability() so that it has the right amount when it makes the request?

FluStudy_au/src/ui/components/GiftCard.tsx
```diff
@@ -58,7 +58,31 @@ class GiftCard extends Component<Props & WithNamespaces, State> {
...
+        this.forceUpdate();
+      }
+    }
+  };
 
+  _checkGiftCardAvailability = async () => {
```
goldjay: Within _checkGiftCardAvailability, I think we need to set the failureReason to "" in the case that the response doesn't have a failureReason. Otherwise if we are out of gift cards and the user reopens the app it will still show the old error.

PR #255, opened 2019-09-26
FEV-1301 Cough giftcard rate limit updates

balderdash:
* Fixes bug where the rate limit check also included demo mode cards that had been assigned in the last day. Now the rate limit applies independently to demo and non-demo cards.
* Add missing await in updateRateLimit
* Add a couple rate limit tests
* Pipe config values into cough_derived so that @smeds1 can add a "approaching rate limit" alert on metabase


tpaik18:
@balderdash While you're in there maybe you can clean out 
`const DEMO_GIFTCARD_URL = "https://www.example.com/giftcard";` which I think is unused now.

PR #275, opened 2019-10-03
Adding FluApi to Fargate

billy-audere:
- Adds a container for nginx with the same configuration we currently use.  Certificate details would be stored in Parameter Store and echoed into files on startup.
- Tries to wire nginx and FluApi as separate containers in a single service.  They should be able to communicate over a loopback address.
- Pulls task execution role into a module since it was a lot of boilerplate.

Breaks metric collection which may need to be solved by adding another sidecar container for stats collection.


billy-audere:
All of the EC2 stuff is still there but I've removed the load balancer attachment from the autoscaling group.  My plan would be to run this only in staging and verify stability before removing the EC2 instance or trying to push to production.

FluApi/docker/nginx/entrypoint.sh
```diff
@@ -0,0 +1,78 @@
...
+  echo_nginx_server "443" "http://127.0.0.1:3000"
+  echo_nginx_server "444" "http://127.0.0.1:3200"
+}
+
+function init_nginx() {
+  rm /etc/nginx/sites-enabled/default
+  echo_nginx_config |  sudo tee "/etc/nginx/sites-enabled/${domain}"
+  # Check the config before restarting
+  nginx -t
+  sudo service nginx restart
```
rightparen: What identity runs this script?

At the top we are creating `/creds`, which should only be possible as root, but here we are calling sudo as though we were not root.

It's probably okay for now if these assume they are running as root or an unrestricted-sudo-capable user.  Eventually it may be nice to write something to `/etc/sudoers.d/50-explicit` to restrict the set of sudo commands allowed to the container user, or see if there is a group we can add the user to that allows systemctl control over nginx.

terraform/modules/flu-api/main.tf
```diff
@@ -104,16 +104,16 @@ resource "aws_route53_record" "api_record" {
...
-  ]
+  #load_balancers = [
+  #  "${aws_elb.flu_api_elb.name}",
+  #  "${aws_elb.flu_api_internal_elb.name}",
+  #]
   max_size = 1
   min_size = 1
   vpc_zone_identifier = ["${var.app_subnet_id}"]
-  wait_for_elb_capacity = 1
+  #wait_for_elb_capacity = 1
```
rightparen: Should all these still be commented out?

FluApi/docker/api/entrypoint.sh
```diff
@@ -9,10 +9,12 @@ umask 077
 SELF_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd)"
 
 time (
+  set +x
   function check() { echo 'select 1;' | psql "$@"; }
   until check "$NONPII_DATABASE_URL" && check "$PII_DATABASE_URL"; do
     sleep 1
   done
+  set -x
```
rightparen: Since this is in a sub-shell, you don't need to `set -x` afterwards.

scripts/manage-service
```diff
@@ -25,45 +25,56 @@ function ecr_login() {
 
 function ecr_publish() {
   local repo="$1"
-  local commit="$2"
+  local tag="$2"
 
   echo "=================================================="
+  echo "Ensuring the destination repo for $repo exists"
+  if [[ -z $(aws ecr describe-repositories | jq '.repositories | .[] | select(.repositoryName == "'"${repo}"'")') ]]; then
```
rightparen: Should quote the entire `"$(...)"` expression.  It shouldn't affect quoting inside.

scripts/manage-service
```diff
@@ -25,45 +25,56 @@ function ecr_login() {
...
+    echo "Creating new repo"
+    aws ecr create-repository --repository-name $repo
```
rightparen: Quote `"$repo"`.

scripts/manage-service
```diff
@@ -25,45 +25,56 @@ function ecr_login() {
...
   echo "=================================================="
-  echo "Re-tagging '$repo:$current_tag' as '$repo:$new_tag'"
-  local manifest="$(aws ecr batch-get-image --repository-name "$repo" --image-ids "imageTag=$current_tag" --query 'images[].imageManifest' --output text)"
-  aws ecr put-image --repository-name "$repo" --image-tag "$new_tag" --image-manifest "$manifest"
+  for repo in "${repos[@]}"
+  do
+    echo "Tagging '$repo:$current_tag' as '$repo:$new_tag'"
+    local manifest="$(aws ecr batch-get-image --repository-name "$repo" --image-ids "imageTag=$current_tag" --query 'images[].imageManifest' --output text)"
+
+    if [[ -z $manifest ]]; then
```
rightparen: Quote `"$manifest"`.

terraform/modules/network/main.tf
```diff
@@ -36,6 +36,15 @@ resource "aws_internet_gateway" "gw" {
...
 }
 
+resource "aws_eip" "nat_gw_eip" {
+  vpc = true
+}
+
+resource "aws_nat_gateway" "nat_gw" {
+  allocation_id = "${aws_eip.nat_gw_eip.id}"
+  subnet_id = "${aws_subnet.app.id}"
+}
```
rightparen: Raised an out-of-band question of whether a public subnet with public IPs but locked-down security groups would work here.

terraform/modules/flu-api/main.tf
billy-audere: The ELB has been deleted.

FluApi/docker/nginx/entrypoint.sh
billy-audere: Yes, currently the nginx container is running as root.

terraform/modules/network/main.tf
billy-audere: NAT gateway removed.

billy-audere:
@rightparen I intended to move this to prod last night but had some trouble setting up the internal load balancer correctly.  A few last points to call out:

- SSH access will be lost to the applications.
- Custom metrics aren't being collected but CPU/mem utilization are captured at the service level, along with per-container logging.  Hopefully our distinct metrics costs will be reduced as a result (the metric dimensions should be simplified to cluster + service) but not sure if we need to delete/reconfigure anything.
- The internal LB to nginx is now HTTP instead of HTTPS (not that the client to the internal LB has always been HTTP).  I couldn't get the LB listener + target group to behave correctly when going from HTTP to HTTPS.  The nginx config now has an SSL + non-SSL server.
- An alarm has been added to monitor that there are active running task instances.  It may make sense to add an alarm to the health check.

billy-audere:
I've also hit a number of deployment issues with the aws_lb when switching target group settings.  Worth noting that generally re-running the deployment resolves the problem.

PR #300, opened 2019-10-18
TFlite RDT reader

jennya:
Adds tflite RDT to Chills for Android. Still a pretty rough prototype, but basically functional:

https://drive.google.com/a/auderenow.org/file/d/1wp4FJgp2xnt9kYPOPo5GLnBlitVrV6OT/view?usp=sharing


FluStudy_us/src/ui/components/flu/AndroidRDTReader.tsx
```diff
@@ -0,0 +1,818 @@
...
+    if (!args.testStripFound) {
+      this._lastRDTReaderResult = rdtCapturedArgsToResult(args);
+      return;
+    }
+
+    const { dispatch, navigation, next } = this.props;
+    dispatch(setRDTCaptureTime(true));
+    try {
+      const photoId = await newUID();
+      const hcPhotoId = await newUID();
```
rightparen: Is high-contrast still relevant?

FluStudy_us/src/ui/components/flu/AndroidRDTReader.tsx
```diff
@@ -0,0 +1,818 @@
...
+      this.setState({
+        fps: ((this._callbackTimestamps.length - 1) / (last - first)) * 1000,
+      });
+      this._callbackTimestamps = this._callbackTimestamps.slice(
+        this._callbackTimestamps.length - 2
+      );
+    }
+  };
+
+  _toggleFlash = () => {
```
rightparen: Do we still need flash?

FluStudy_us/src/ui/components/flu/AndroidRDTReader.tsx
jennya: We don't do anything to make it high-contrast, but I am using the test area bitmap here which we can display to the user on a later screen to draw attention to the part of the test strip that we care about. I don't think this is necessarily a path we want to go down, but thought it was trivial to implement so left it in for David/Shawna to evaluate.

FluStudy_us/src/ui/components/flu/AndroidRDTReader.tsx
jennya: Maybe not -- will commit as is for now but discuss removing it when we meet with David later today

PR #301, opened 2019-10-18
FEV-1317 Intro screens to Importance of Accuracy screen

audererob:
[FEV-1317](https://auderenow.atlassian.net/browse/FEV-1317)
This change includes the following UX changes:
* Modify AnimatedChrome and Chrome to not draw the splash image, and only draw the background "bug"
* Add new introductory screens
* Change coloring of introductory screens to purple background w/white text
* Change "Continue" button to hyperlink-style
* Enlarge font size
* Allow text style to be set (for changing color) on BulletPoint and CollapsibleText
* Allow left-justifying title text


FluStudy_us/src/i18n/locales/en.json
```diff
@@ -168,8 +168,32 @@
     "seeExample": "Tap the \"?\" below to see examples"
   },
   "Welcome": {
-    "title": "Welcome",
-    "desc": "We know you aren't feeling well right now. Thank you for participating even when you're sick. With your help, we want to understand if people (like you) can test themselves at home for the flu."
+    "title": "Why we are doing this",
```
goldjay: Does this need a question mark?

audererob: I'll double-check with Shawna but the Jira doesn't have one.

FluStudy_us/src/i18n/locales/en.json
```diff
@@ -168,8 +168,32 @@
...
+    "desc": "We know you aren't feeling well right now. Thank you for participating even when you're sick.\n\nWith your help, we want to understand if people (like you) can test themselves at home for the flu. But first, we'll walk you through a few things:",
+    "desc2": "How does the test work and what's required?\nHow am I helping with medical research?\nWhat can I expect to learn?"
+  },
+  "HowDoesTestWork": {
+    "title": "How does the test work?",
+    "desc": "You're about to do two tests on yourself, each involving a nasal swab. Don't worry, it shouldn't be too uncomfortable. This app will guide you through each step.\n\nIt's important for our study that you complete the test as **carefully** as possible. In addition, please answer each question as **honestly** as possible. Our research, and ability to give you a result after the test depends on accurate data, so we thank you in advance."
```
goldjay: The comma placement is a little strange in the last sentence: "Our research, and ability to give you a result after the test depends on accurate data, so we thank you in advance." Maybe no commas?

FluStudy_us/src/i18n/locales/en.json
audererob: Actually it's a statement - I keep reading it as "why are we" but it's "why we are".

FluStudy_us/src/i18n/locales/en.json
audererob: Agreed - that's what's in the Jira but it's worth editing.

FluStudy_us/src/ui/components/EmailEntry.tsx
```diff
@@ -0,0 +1,95 @@
...
+            style={styles.textInput}
+            value={this.state.email}
+            onChangeText={this._onEmailChange}
+          />
+        </View>
+      </KeyboardAvoidingView>
+    );
+  }
+
+  _validEmailShape(email: string): boolean {
```
goldjay: You can just copy/paste the email validation from archive/flustudy_us

FluStudy_us/src/ui/components/flu/TestResult.tsx
```diff
@@ -71,11 +71,11 @@ class TestResult extends React.Component<Props & WithNamespaces> {
         <View style={{ marginHorizontal: GUTTER }}>
           <BulletPoint
             content={t("common:testResult:blueLine")}
-            customBulletUri="listarrow"
+            customBulletUri="bullet"
```
goldjay: I may be remembering incorrectly, but isn't bullet the default for BulletPoint? If you want to change the size/shape of the bullet it may be nicer to just change the default.

FluStudy_us/src/ui/components/EmailEntry.tsx
audererob: Awesome, great idea!

FluStudy_us/src/ui/components/flu/TestResult.tsx
audererob: It isn't the default, it uses a unicode character as that. But since nothing uses the default, it's a good idea you have to make bullet the default and remove a bunch of customBulletUri properties in screenconfig.

PR #302, opened 2019-10-21
FEV-1352 Keyboard obscures email input field

audererob:
[FEV-1352](https://auderenow.atlassian.net/browse/FEV-1352)
The KeyboardAvoidingView in the EmailEntry component wasn't working properly, so moved it out as a capability in ScreenConfig. Updated BarcodeEntry accordingly.

Also contains a bonus tweak of fixing the line height for the smaller text on the "Did you know?" screen.


FluStudy_us/src/ui/components/Screen.tsx
```diff
@@ -232,33 +238,50 @@ export const generateScreen = (config: ScreenConfig) => {
...
-              styles.scrollContainer,
-              { backgroundColor: config.backgroundColor || "white" },
-            ]}
-          >
-            <CustomScrollView
-              contentContainerStyle={styles.contentContainer}
-              keyboardShouldPersistTaps="handled"
-              bounces={!disableBounce}
+          {!!config.keyboardAvoidingView ? (
+            <KeyboardAvoidingView
```
goldjay: Nice! It makes sense to put this around the screen since there is a change that, at the component-level, it may not cover the whole screen.

PR #303, opened 2019-10-21
Allowing for multiple data pipelines within the same database.

billy-audere:
Creates another table to track pipelines for external reference. Also refactors the pipeline code a bit in attempt to make various pieces more reusable.

Specifically the Firebase nodes will probably be reused.  And the pipeline service code should be decoupled from the Cough-specific stuff.


FluApi/src/services/data/dataPipeline.ts
```diff
@@ -0,0 +1,150 @@
+// Copyright (c) 2019 by Audere
+//
+// Use of this source code is governed by an MIT-style license that
+// can be found in the LICENSE file distributed with this file.
+
+import { Sequelize } from "sequelize";
+
+export interface DataPipeline {
+  name: string;
+  db: Sequelize;
```
rightparen: NBD for now, but it doesn't seem quite right for the pipeline itself to keep this.  E.g. I expect we will have pipelines that span data sources.  I'd expect the update method to bring together a DataPipeline and some context that could provide access to the relevant data sources.

That said, I'm fine if you prefer to change this later once we have a scenario.

billy-audere: Yes, for the time being it seemed less cluttered to have this at the task/node level only because we have a single database.  That said it probably fits better there as we may want to interact with multiple data sources.

It would be worth refactoring at some point if the current pipeline use cases are expanded.

PR #307, opened 2019-10-22
Adding new timer UX, and content changes

audererob:
Changing up the timer UX as per recent designs, and also changing some copy per the most recent updates.
Screen Shot 2019-10-22 at 3 04 36 PM
Screen Shot 2019-10-22 at 3 05 57 PM


audererob:
I like the idea of hiding Continue; the design had it be disabled so I went with that.

PR #310, opened 2019-10-23
terraform: add IPRD RDT-reader server in ECS

rightparen:
To support this change, I had to make our nginx image support SUBPATH and SERVERS variables.

SUBPATH is the URL sub-path to map to the service container.  E.g. for the RDT reader this is "/IPRD/rdt-reader/".

SERVERS specifies how to listen.  It is a space-separated list of entries where each entry looks like "protocol:inport:outport" (without the quotes).  For example, the entry "https:443:3000" species to listen on "https" port "443" and redirect to a service on port "3000".


terraform/modules/flu-api/main.tf
```diff
@@ -226,3 +227,118 @@ resource "aws_cloudwatch_metric_alarm" "fluapi-task-alarm" {
...
+  desired_count = 1
+  deployment_maximum_percent = 200
+  deployment_minimum_healthy_percent = 100
+  launch_type = "FARGATE"
+  network_configuration {
+    assign_public_ip = true
+    security_groups = [
+      "${var.internet_egress_sg_id}",
+      "${var.fluapi_server_sg_id}",
+      "${var.db_client_sg_id}",
```
billy-audere: This may be extraneous.

terraform/modules/flu-api/main.tf
```diff
@@ -226,3 +227,118 @@ resource "aws_cloudwatch_metric_alarm" "fluapi-task-alarm" {
...
+      "${var.dev_ssh_server_sg_id}",
```
billy-audere: We should probably get rid of this SG for the Fargate tasks.

rightparen:
Note that this does not fully work yet--this is just a checkpoint that gets nginx working for this scenario and implements a scaffold for the RDT reader service.

The current failure is that nginx reports connection refused from the service container:
```
2019/10/23 20:29:03 [error] 18#18: *3 connect() failed (111: Connection refused) while connecting to upstream, client: 10.0.254.42, server: api.staging.auderenow.io, request: "POST /IPRD/rdt-reader/align HTTP/1.1", upstream: "http://127.0.0.1:9000/align", host: "api.staging.auderenow.io"
```
I suspect the problem might be that the image I have for the RDT reader listens on `http://0.0.0.0:9000` rather than `http://127.0.0.1:9000`, but I am currently blocked on producing a new image by another issue.

PR #314, opened 2019-10-24
FEV-1358: clean up firebase logging

smeds1:
I removed firebase logging events that I don't believe anyone is using. Ideally this change will significantly cut down the number of events we log per survey and speed up firebase queries.


FluStudy_us/src/util/fluResults.ts
```diff
@@ -76,19 +76,6 @@ export async function logFluResult() {
   _previousPinkAnswer = pinkAnswer;
 }
 
-export async function logNumLines() {
-  const state = (await getStore()).getState();
-
-  const numLinesAnswer = getSelectedButton(state, NumLinesSeenConfig);
-  if (_previousNumLines && _previousNumLines !== numLinesAnswer) {
-    logFirebaseEvent(FunnelEvents.NUM_LINES_ANSWER_CHANGED, {
```
rightparen: Is this not interesting?  I thought we had an insight from this.

smeds1: @rightparen We don't have a high contrast image in Chills, so the user doesn't answer the number of lines question anymore.

FluStudy_us/src/ui/ConnectedRootContainer.tsx
```diff
@@ -282,8 +282,6 @@ class ConnectedRootContainer extends React.Component<Props & WithNamespaces> {
     switch (action.type) {
       case NavigationActions.NAVIGATE:
       case StackActions.PUSH:
-        return NavEvents.FORWARD;
-
```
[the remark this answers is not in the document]
smeds1: @tpaik18 I find it interesting when a user navigates backward. Knowing that they want to revisit content seems useful. Seeing that a user navigates forward doesn't really provide any insight. 

I think you're right that I didn't properly remove FORWARD from the switch statement. Thanks for pointing that out.

FluStudy_us/src/ui/ConnectedRootContainer.tsx
```diff
@@ -282,8 +282,7 @@ class ConnectedRootContainer extends React.Component<Props & WithNamespaces> {
...
+        break;
```
audererob: If those two events are going to no-op (i.e. return undefined), then you can just remove the two cases in the switch entirely.

FluStudy_us/src/util/fluResults.ts
```diff
@@ -76,19 +76,6 @@ export async function logFluResult() {
...
-      old_answer: _previousNumLines,
-      new_answer: numLinesAnswer,
-    });
-  }
-  _previousNumLines = numLinesAnswer;
-}
-
```
audererob: You can also remove the declaration of _previousNumLines at the top of the file.

FluStudy_us/src/ui/ConnectedRootContainer.tsx
smeds1: Thanks, @audererob

PR #322, opened 2019-10-28
Enabling import of Chills survey documents from Firebase into AWS.

billy-audere:


FluApi/src/endpoints/webPortal/auth.ts
```diff
@@ -179,6 +179,10 @@ export enum Permission {
   COUGH_RDT_PHOTOS_ACCESS = "coughRdtPhotosAccess",
   COUGH_RDT_PHOTOS_WRITE = "coughRdtPhotosWrite",
   COUGH_INTERPRETATION_WRITE = "coughInterpretationWrite",
+  CHILLS_GIFTCARD_UPLOAD = "chillsGiftcardUpload",
```
smeds1: Do we have gift cards for Chills?

FluApi/src/models/db/chills.ts
```diff
@@ -0,0 +1,306 @@
...
+      newPhotoHash: stringColumn("newPhotoHash"),
+      replacerId: unique(integerColumn("replacerId")),
+    },
+    { schema }
+  );
+}
+
+// ---------------------------------------------------------------
+
+export interface PiiReviewAttributes {
```
smeds1: My impression is that we don't have to worry about PII showing up in photos anymore since we're not using clinic barcodes.

FluApi/src/app.ts
```diff
@@ -37,6 +37,7 @@ import { ServerHealth } from "./endpoints/healthCheck";
 import { CoughFirebaseEndpoint } from "./endpoints/coughFirebaseApi";
 import { CoughFollowUpEndpoint } from "./endpoints/coughFollowUpApi";
```
smeds1: Do we need this now that there is no follow-up survey?

billy-audere: There is still a follow up survey for Cough administered through Qualtrics.

FluApi/src/models/db/chills.ts
billy-audere: Can we get rid of PII review for photos?  I figured someone might inadvertently put something with their name or whatever in the photo.  I'm honestly not that familiar with the use cases.

FluApi/src/app.ts
smeds1: @billy-audere Sorry, tried to review too fast. I thought this line was for Chills.

FluApi/src/models/db/chills.ts
smeds1: Shawna says we can get rid of the PII review.

FluApi/db/migrations/nonpii/20191022105323-create-chills-tables.js
```diff
@@ -0,0 +1,114 @@
+// Copyright (c) 2019 by Audere
+//
+// Use of this source code is governed by an MIT-style license that
+// can be found in the LICENSE file distributed with this file.
+
+"use strict";
+
+const { baseColumns, column, unique, foreignIdKey } = require("../../util");
+const schema = "chills";
+const derivedSchema = "chills_derived";
```
rightparen: These seem like tables that should go in `"chills"` rather than `"chills_derived"`.

billy-audere: The tables in this migration should all be going into `"chills"`.  The `"chills_derived"` schema is being created but not populated.

PR #336, opened 2019-10-31
Take high res RDT image when filters pass

jennya:
Adds logic to take a high resolution photo (highest resolution available on the device) once the phase 1 and image quality filters pass. Had to do some substantial refactoring of the ImageListener because in order to re-run the filters on the high resolution image which has a larger size, we need to transform that image into the format the tflite model is expecting and adjust the size parameters given to the RDTTracker


jennya:
One major optimization -- encoding the high quality image into base64 was unimaginably slow. This led me to save the jpeg directly into internal storage (available only to the app) and then provide the file uri to RN for display.

We still need to send that jpeg to Firebase since we are not able to upload the base64 image in RN. Not sure if that logic should be in Android or RN.

PR #344, opened 2019-11-04
Check in stub wrapper for IPRD filter

rightparen:
I got the IPRD filter code working locally and removed their code and commented out enough of the wrapper that we still compile and run.


FluStudy_us/android/app/src/main/java/host/exp/exponent/DetectorView.java
```diff
@@ -355,28 +379,43 @@ private void processImage() {
...
+                                Log.i(TAG, "IPRD processing time: " + (SystemClock.uptimeMillis() - iprdStartTimeMs) + "ms");
 
-                            final List<Classifier.Recognition> results = boxDetector.recognizeImage(boxModelBitmap);
-                            Log.i(TAG, "Phase 1 processing time: " +  (SystemClock.uptimeMillis() - boxStartTimeMs) + "ms");
+                                // Local interpretation prototype
+                                Trace.beginSection("Running Process Image");
+                                final long boxStartTimeMs = SystemClock.uptimeMillis();
 
-                            final List<Classifier.Recognition> mappedRecognitions = filterResults(BOX_MINIMUM_CONFIDENCE_TF_OD_API, results, true);
+                                final List<Classifier.Recognition> results = boxDetector.recognizeImage(boxModelBitmap);
```
jennya: Should we run our classifier if the !iprdResult.isAccepted() ? Seems like maybe we should just skip ours in this case

rightparen:
@jennya that's fair.  I had originally thought we might want statistics on images that pass one but not the other, but that is probably not interesting.  I'll remove.

jennya:
Ah yes that probably is interesting but would come at a notable performance hit on low-end devices. I wonder if we could do a check for whether we're currently on a device where we think the affect would be minimal and then run both?

PR #350, opened 2019-11-05
FEV-1138 Updated questionnaire and supporting UX

audererob:
* Added DatePicker to choose a date (not to be confused with the existing MonthPicker)
* Added MultiDropDown to choose countries visited
* New SurveyQuestionTypes - date, multi-dropdown, zipcode
* Input questions can have placeholder text
* Added text variables to question text
* Various styling updates

Bonus fix to timer fast forward functionality: triple-clicking would typically count "5, 3, 2, 1". This was due to the start time getting set to exactly 5s remaining, and then the system timer being reset following that - upon ticking, the time remaining would get computed at ~3.99 and Math.floor would make that 3s remaining. The fix is to add more than 5s but less than 6s (choosing 5.9s for this change) to account for any slowness between setting the remaining time and (re)starting the system timer.


lib/chillsQuestionConfig.ts
```diff
@@ -1119,59 +812,44 @@ export const NumLinesSeenConfig: SurveyQuestion = {
...
+  PeopleInHouseholdConfig,
+  ChildrenAgeGroupsConfig,
+  ChildrenDaycarePreschoolConfig,
+  SomeoneDiagnosedConfig,
+  InContactConfig,
+  PublicTransportationConfig,
+  AroundSickChildrenConfig,
   BlueLineConfig,
   PinkWhenBlueConfig,
   PinkLineConfig,
```
smeds1: Obsolete?

FluStudy_us/src/i18n/locales/en.json
```diff
@@ -125,51 +128,437 @@
...
+    "sixOrMore": "6 or more",
+    "noPink": "Blue line only",
+    "yesAboveBlue": "Pink line above the blue line",
+    "yesBelowBlue": "Pink line below the blue line",
+    "yesAboveBelowBlue": "Pink line above and below the blue line",
+    "yesOnePink": "Yes, there is one pink line",
+    "yesTwoPink": "Yes, there are two pink lines",
+    "oneLine": "One line only",
+    "twoLines": "Two lines",
+    "threeLines": "Three lines"
```
smeds1: I believe these are also obsolete

audererob: They are indeed obsolete - I'll leave these as-is for now since they have tendrils into a decent amount of other code, and would bloat this already large PR. When we scrub through the various "lines" questions and code we'll take care of it then.

FluStudy_us/src/ui/components/Questions.tsx
```diff
@@ -139,16 +152,29 @@ class Questions extends React.PureComponent<Props, State> {
...
+            highlighted={highlighted}
+            question={config as TextQuestion}
+          />
+        );
+      case SurveyQuestionType.ZipCodeInput:
+        return (
+          <NumberInputQuestion
+            highlighted={highlighted}
+            maxDigits={5}
+            question={config as TextQuestion}
```
tpaik18: It's a little odd that ZipCodeInput is just a NumberInput with max digits 5. Does this mean that any four-digit number is okay?

FluStudy_us/src/ui/components/TimerRing.tsx
```diff
@@ -17,7 +17,7 @@ import MultiTapContainer from "./MultiTapContainer";
 const SECOND_MS = 1000;
 const SECONDS_MINUTE = 60;
 const MINUTE_MS = SECONDS_MINUTE * SECOND_MS;
-const FAST_FORWARD_MS = 5 * SECOND_MS;
+const FAST_FORWARD_MS = 5.9 * SECOND_MS;
```
tpaik18: Can we do maybe 5.1? It's nice not to wait an extra 0.9 seconds.

FluStudy_us/src/ui/components/Questions.tsx
audererob: I believe you're correct - I'll add a minDigits to it.

FluStudy_us/src/ui/components/TimerRing.tsx
audererob: It doesn't affect the wait time since Math.floor is used to compute whether the timer is done. However I can make it 5.1 to not feel quite so cruddy.

PR #365, opened 2019-11-08
FEV-1277 FEV-1151 FEV-1146 FEV-1132 Content updates and plumbing Achievement account info

audererob:
Various content updates for UTM test and shipping, and plumbing in patient's Achievement info along with email verification logic. It should be fairly straightforward to integrate with the server API call once it becomes available.

Also added the ability for embedded links to provide an argument of sorts, allowing for the user's state to be passed through to the FedEx URL.

Also trying possible fix to FEV-1406.


FluStudy_us/src/ui/components/NumberInputQuestion.tsx
```diff
@@ -16,14 +16,31 @@ import NumberInput from "./NumberInput";
...
+  Props & WithNamespaces,
+  State
+> {
+  constructor(props: Props & WithNamespaces) {
+    super(props);
+    this.state = {
+      text:
+        this.props.answer !== undefined
+          ? this.props.minDigits
+            ? this.props.answer.toString().padStart(this.props.minDigits, "0")
```
goldjay: TIL about padStart

FluStudy_us/src/util/patientAchievementInfo.ts
```diff
@@ -0,0 +1,66 @@
...
+  const { actualEmail } = state.meta.patientAchievementInfo;
+  const atIndex = actualEmail.indexOf("@");
+  if (atIndex >= 0) {
+    const start = atIndex > 1 ? actualEmail[0] : FILLER_CHAR;
+    const middle = atIndex > 2 ? FILLER_CHAR.repeat(atIndex - 2) : "";
+    const end =
+      atIndex > 3 ? actualEmail[atIndex - 1] : atIndex > 1 ? FILLER_CHAR : "";
+    emailHint = start + middle + end + actualEmail.substring(atIndex);
+  } else {
+    // Error??
```
goldjay: This is the case in which we have an invalid email address without an "@". Is this possible at this point? Also, would it make sense for atIndex > 0 since I don't think it's possible to have an email that is just something like "@gmail.com"

audererob: I don't think the bad email should be possible - we can assume the server will return back sane data. I'll take out that else case.

PR #366, opened 2019-11-08
FEV-1411: replace cough survey questions with chills questions

smeds1:
This PR is an attempt to fix Chills automation for Sunflower after the new survey questions were introduced. Things that still need to be done in later PRs:

* Update the code to run on Android as well as iOS
* Put in the ability to pick specific dates (right now every date entry defaults to the current date)
* Put in the ability to pick one or more countries from the dropdown. I'm going to wait and see if that UI gets modified any more before seeing if Appium can handle the question.
* Put in the ability to pick city/state/zip for where the majority of time was spent. This set of questions seems like it should be easy to implement, but I was getting unexpected crashes on my first attempt.
* Uncomment the question about daycare/preschool (waiting for FEV-1406 fix)
* Add one more script so that all choices can be exercised (there is at least one question with 7 choices)


FluStudy_us/test/uiTests/testInputs/default.js
```diff
@@ -6,15 +6,14 @@
 import strings from "../../../src/i18n/locales/en.json";
 
 export const inputs = {
-  [strings.barcode.placeholder]: "00000000",
-  [strings.barcode.secondPlaceholder]: "00000000",
-  [strings.EmailConfirmation.placeholder]: "joan@auderenow.org",
+  [strings.barcode.placeholder]: "0000000000",
+  [strings.common.emailEntry.placeholder]: "joan@philip@auderenow.org.org",
```
tpaik18: Intentionally malformed?

smeds1: Thanks for catching. I changed the email at the last minute after Rob's PR went in.

PR #369, opened 2019-11-12
Fix three typos in Chills copy

tpaik18:
The "Achievment" typo was not from the copy doc. 
The other two typos are from the copy doc, and I have made suggested edits in the copy doc to match.


FluStudy_us/src/i18n/locales/en.json
```diff
@@ -811,7 +811,7 @@
   },
   "MucusUTM": {
     "title": "Collect sample from nose",
-    "desc": "Now you'll swab your nose the same was you did in the first test.",
+    "desc": "Now you'll swab your nose the same as you did in the first test.",
```
smeds1: I think this was supposed to be "way"

FluStudy_us/src/i18n/locales/en.json
```diff
@@ -841,7 +841,7 @@
   },
   "WebSiteForPickup": {
     "title": "Schedule a FedEx pickup online",
-    "desc": "You can fill out a form on the FedEx website to arrange for a pickup. If you leave the package on your doorstep for pickup, try not leave it in direct sunlight or freezing temperatures.\n\nWhen filling out the form on the FedEx site, please note:",
+    "desc": "You can fill out a form on the FedEx website to arrange for a pickup. If you leave the package on your doorstep for pickup, try to not leave it in direct sunlight or freezing temperatures.\n\nWhen filling out the form on the FedEx site, please note:",
```
smeds1: I like "not to" better than "to not"

FluStudy_us/src/i18n/locales/en.json
tpaik18: @smeds1 Oh!

FluStudy_us/src/i18n/locales/en.json
tpaik18: @smeds1 Me too, and I first did "not to" but then I saw it was "to not" in the other screen that has the same sentence.

tpaik18:
Shawna just approved some of my edits on a few other screens so I am rolling those fixes into this PR.

PR #370, opened 2019-11-12
Remove progress circle, fix over exposure bug

jennya:
Found that the preview frame maxWhite was often 255.0 which equals our threshold, changed test to be > threshold rather than >= for OVER_EXPOSED

Removed unused progress circle since it's not updated anymore


rightparen:
Interesting.  I'm curious about the 255.0 value--are the values coming from bytes, and if so, is it even possible to get a value >255?

PR #375, opened 2019-11-13
Add title to barcode template

smeds1:
I was still seeing "flu@home Australia" as the title on the Chills barcode review page. I added a title to the template, which should hopefully fix the problem. I wasn't able to test locally because I wasn't sure how to set the local permissions to enable RDT photo viewing.


smeds1:
Thanks @balderdash

PR #376, opened 2019-11-13
Offline Barcode Verification + Back Beyond Barcode Verification Reset

goldjay:
[FEV-1251](https://auderenow.atlassian.net/browse/FEV-1251)
The backend portion for barcode verification isn't quite ready yet and the required information may change to involve some sort of encryption so the actual method call is commented out.

* If you get to the barcode verification screen and you are offline, it should send you to a screen saying you need to be connected to verify your barcode with a button that tries verification again.

* If you go back beyond the barcode verification screen, an alert will pop up asking if you would like to reset the app. If you choose "yes", it should clear all of the data and start you at the welcome screen.


FluStudy_us/src/ui/components/NavigationBar.tsx
```diff
@@ -46,6 +51,10 @@ class NavigationBar extends React.Component<Props & WithNamespaces> {
       return;
     }
 
+    if (!!this.props.onBack) {
+      this.props.onBack(this.props.navigation, this.props.dispatch);
```
audererob: I think we'll want a "return;" here to prevent the user from backing up past the email confirmation screen.

FluStudy_us/src/ui/components/NavigationBar.tsx
```diff
@@ -46,6 +51,10 @@ class NavigationBar extends React.Component<Props & WithNamespaces> {
...
+    }
+
     // @ts-ignore
     this._debounce = setTimeout(() => {
```
audererob: Unrelated to this PR: oof, I've never noticed that code before. We should totally consider getting rid of it unless it's required?

FluStudy_us/src/ui/components/NavigationBar.tsx
goldjay: If we return here, the user will never be able to go back beyond that screen, even if they select "no" on the alert. I could wrap the result in a promise that would resolve/reject based on their answer. Do you think awaiting the result of the alert would mess with navigation? I have a vague memory of someone doing async operations during navigation before.

FluStudy_us/src/ui/components/NavigationBar.tsx
goldjay: Do you mean the debounce method? I'm pretty sure that's needed to prevent people from getting ahead of the navigation stack.

FluStudy_us/src/ui/components/NavigationBar.tsx
audererob: I think preventing the user from going back from that screen when they say "No" to resetting is exactly what we want - @ShawnaCooper @tpaik18 would you agree?

FluStudy_us/src/ui/components/NavigationBar.tsx
audererob: Yeah, the debounce - I know folks have complained that you can't go back very quickly. If it would cause problems to remove completely then maybe we can play around with lowering the debounce time in the nearish future.

ShawnaCooper:
Just to be sure I'm following this - it's the process we discussed before
where we ask people if they want to reset if they try going back past the
email verification screen. Right?

Thanks.
-Shawna

goldjay:
@ShawnaCooper Yes, that's the one.

ShawnaCooper:
Great. Then I agree! Thanks.

FluStudy_us/src/ui/components/NavigationBar.tsx
tpaik18: @goldjay @audererob Agree. If they say No to resetting, then don't go back, just do nothing.

goldjay: Okay, I'll add the "return" in an upcoming barcode verification PR.

tpaik18: Thanks @goldjay 
Also nit: Maybe we can get rid of resetTimestamp() in src/store/survey.ts; there appear to be no references to that function now.

PR #378, opened 2019-11-13
FEV-1397 Capture install referrer on Android

tpaik18:
Using react-native-referrer + some of @auderephilip's old code in https://github.com/AudereNow/learn/pull/736 to pass the `referrer` in the Play Store download link through to the survey JSON (specifically, we write this to the existing but currently unused `marketingProperties` field). 

Used the instructions in https://developers.google.com/analytics/solutions/testing-play-campaigns#broadcast to broadcast the intent locally via adb, and verified that this gets written to the survey: 
Screen Shot 2019-11-13 at 2 27 06 PM


FluStudy_us/android/app/build.gradle
```diff
@@ -127,7 +127,8 @@ apply from: 'expo.gradle'
 apply from: "../../node_modules/react-native-unimodules/gradle.groovy"
 
 dependencies {
-    implementation project(':react-native-modal-translucent')
+  compile project(':react-native-referrer')
```
auderephilip: Do you get warnings that `compile` is deprecated?  My understanding is that everyone should be migrating towards `implementation`.

tpaik18: TIL, thanks.

tpaik18:
Just to be clear, this feature works on Android only. The consensus on the internet seems to be No Can Do on iOS.

PR #391, opened 2019-11-14
FEV-1444: eliminate US from foreign countries

smeds1:
When asked what countries they traveled in outside the U.S., the user should not be able to pick the U.S.


smeds1:
Sorry about that. I didn't realize that whole list was written out twice. This version works.

tpaik18:
Okay, two more suggestions are `yarn build` and also rename COUNTRY_MULTIDROPDOWN_DATA to FOREIGN_COUNTRY_MULTIDROPDOWN_DATA just so it's clear to everyone that this is an official list of countries minus the US?

PR #392, opened 2019-11-14
Barcode validation crypto

goldjay:
[FEV-1133](https://auderenow.atlassian.net/browse/FEV-1133)

MISC Changes:
* Replaced "*" with "_" in email hint since every 4 asterisks count as trying to bold the inner text
* Changed 
* Added a check for csruid on navigation since, after the state was cleared via alert dialogue, it was never set again


FluStudy_us/src/util/patientAchievementInfo.ts
```diff
@@ -31,43 +39,67 @@ const DEMO_BARCODE = "1111111111";
...
-  const atIndex = actualEmail.indexOf("@");
-  if (atIndex > 0) {
-    const start = atIndex > 1 ? actualEmail[0] : FILLER_CHAR;
-    const middle = atIndex > 2 ? FILLER_CHAR.repeat(atIndex - 2) : "";
-    const end =
-      atIndex > 3 ? actualEmail[atIndex - 1] : atIndex > 1 ? FILLER_CHAR : "";
-    emailHint = start + middle + end + actualEmail.substring(atIndex);
-  }
+  let { email } = state.meta.patientAchievementInfo;
+  email = email.split("*").join("_");
```
audererob: FYI, if you make "_" instead be FILLER_CHAR then it will render properly as an asterisk.

FluStudy_us/src/util/patientAchievementInfo.ts
```diff
@@ -31,43 +39,67 @@ const DEMO_BARCODE = "1111111111";
...
+          .pbkdf2Sync(
+            state.meta.enteredEmail,
+            state.meta.patientAchievementInfo.emailSalt,
+            10000,
+            64,
+            "sha512"
+          )
+          .toString("base64");
+  if (state.meta.isDemo) {
+  }
```
audererob: No longer needed?

PR #394, opened 2019-11-14
FEV-1420 and don't show required hint all over

tpaik18:
Sadly, I couldn't do the logic of Questions showing RequiredHint when !!showRequiredHint because David wants to keep the dividers, and then the required hint would show below the divider, instead of above the divider. 

So I reverted Questions back to what it was before, and now explicitly added tag: RequiredHint above Divider into screens where we want to show it (currently just the first one). 

Also using sentence casing for all survey screens as David suggested.


tpaik18:
@smeds1 Correct, reverting to how it was before I started my crusade to turn the asterisk red.

PR #396, opened 2019-11-15
FEV-1251 Change screen where prompt shows and prevent navigating back

audererob:
[FEV-1251](https://auderenow.atlassian.net/browse/FEV-1251)
Changing the screen where we pop the alert asking the user whether they want to use a different kit - formerly this was "Verify your email" but now is to be "Set up kit box". Also allowing prevention of back navigation so the user can't reach the "Verify your email" screen.


audererob:
Correct, navigation previous to the "set up your kit box" screen will not be allowed.

PR #400, opened 2019-11-18
Import Chills kit orders

billy-audere:
- Adds config for Google API specific secrets.  Notably this includes a refresh token for an OAuth Client ID that has been granted access to the Google Sheets in question.  Currently I am the authorizer, which may impact client access if my authority is revoked.
- The intention is to have a separate OAuth ID for staging and production.  However as best I can tell access can only be granted to all sheet which I have permission to.  So each client would have indistinguishable permissions.  It may be possible to create new accounts for each sheet and grant permission that way.
- Retrieves non-demo data from a given sheet and demo data from a separate sheet.  This will allow QA to manage demo records manually.
- Overwrites both demo & non-demo records on import.
- Validates barcodes & states on import.  If these are invalid the import will fail.
- Adds Evidation id to the table in `chills_derived` and refreshes after import.  More fields could be added but most of the kit information we have is either PII (email) or sensitive.
- Creates a Lambda job for invoking the import every weekday.  The Google Sheet will likely be updated manually by Evidation.

I have performed functional testing with a valid token.  Need to write or perform some additional tests before merge but this should be good to review as I wouldn't anticipate much change due to testing.


FluApi/src/endpoints/chillsImportKits.ts
```diff
@@ -0,0 +1,73 @@
...
+  constructor(sql: SplitSql) {
+    const secrets = new SecretConfig(sql);
+    this.config = new LazyAsync(async () => getGoogleApisConfig(secrets));
+
+    this.googleClient = new LazyAsync(async () => {
+      const apiConfig = await this.config.get();
+      const client = new google.auth.OAuth2({
+        clientId: apiConfig.clientId,
+        clientSecret: apiConfig.clientSecret,
+        redirectUri: "https://localhost.com",
```
rightparen: Interesting.  I'm surprised we don't need to exercise this to get results back if auth works.

billy-audere: This fragment was missing from the PR on the following line, which may explain:

```
      client.setCredentials({
        refresh_token: apiConfig.refreshToken,
      });
```

The authorization workflow isn't used by the application because creating a new token generally requires explicit user approval.  A refresh token is posted to the token endpoint along with the client id & secret to get a new access token without user input.  All of that setup occurs before using the API, it is likely not ideal but I created an initial access token with an offline (refresh) grant that has permission to act on my behalf.

PR #403, opened 2019-11-19
Remove unnecessary thread.join() calls in onPause FEV-1363

jennya:
When backgrounding the app, occasionally (reliably reproduces ~20% of the time) the join() call in the CameraController would block and prevent the RDT reader from successfully exiting. This causes the app to freeze on re-open. I don't know why quitSafely() wasn't causing the backgroundThread to actually quit, but the join was enforcing a synchronization event that actually doesn't seem necessary. Removing the .join() call prevents the freeze behavior. @rightparen does this seem ok to you?


jennya:
Using the profiler to investigate what the background thread was doing after the quitSafely() call wasn't helpful -- the thread state under normal conditions was either "Running", "Sleeping" or "Dead", under this condition it was "Unknown"

FluStudy_us/android/app/src/main/java/host/exp/exponent/customview/OverlayView.java
```diff
@@ -1,37 +0,0 @@
...
-
-package host.exp.exponent.customview;
-
-import android.content.Context;
-import android.graphics.Canvas;
-import android.util.AttributeSet;
-import android.view.View;
-
-/** A simple View providing a render callback to other classes. */
-public class OverlayView extends View {
```
rightparen: Why is this class no longer needed after changing `thread.join`?

jennya: It's not related. Just removing an unused class

jennya:
I can see in the profiler that the thread was previously active and doing work (cycling between running and sleeping). I will create a jira task to track this -- since we're using quitSafely(), it's possible that there are some messages still in the queue for this thread that sometimes can't complete

PR #404, opened 2019-11-19
Add FirebaseStorageUploadModule on Android

rightparen:
We want the ability to upload files without converting them to base64 strings.  The existing react native firestore libraries only allow uploads of in-memory strings.  Also, the current implementation is a bit complicated.

This factors the upload process into a Javascript orchestrator and a native data manager.  The Javascript orchestrator is based on the existing PhotoUploader, can track opportune times to upload based on UI idleness, and can leverage existing network state change tracking logic.  The native data manager (only in Android so far) manages a directory of files pending upload and uploads them to Firebase Storage.

iOS (and `resultWindowImgBase64` -> `resultWindowImageUri`) is not yet implemented.


FluStudy_us/src/store/index.ts
```diff
@@ -73,8 +73,12 @@ const photoUploader = new PhotoUploader({
   collection: photoCollectionName(),
 });
 
-export function savePhoto(photoId: string, jpegBase64: string) {
-  return photoUploader.savePhoto(photoId, jpegBase64);
+export function uploadFile(uid: string, filepath: string) {
+  return photoUploader.enqueueFileContents(uid, filepath);
+}
+
+export function uploadBase64String(photoId: string, base64String: string) {
```
jennya: Can we remove this?

rightparen:
Okay, I can remove the result window image path when I implement the iOS version of this.

PR #405, opened 2019-11-19
FEV-1477 Use device-specific noun instead of "device" 

tpaik18:
Bringing back the feature we had in Fever of dynamically showing "iPhone", "tablet", "phone" etc based on the hardware we were running on. Talked about this a few weeks ago with David, @jennya, @rightparen in an RDT design meeting, then I went back to the code and saw it needed some plumbing to get the variable passed through. Here's what we will show: 

**iOS phone -> "iPhone"** 
Simulator Screen Shot - iPhone 8 - 2019-11-18 at 15 31 02
Simulator Screen Shot - iPhone 8 - 2019-11-18 at 16 00 53

**iOS tablet -> "iPad"** 
Simulator Screen Shot - iPad (6th generation) - 2019-11-18 at 16 03 03
Simulator Screen Shot - iPad (6th generation) - 2019-11-18 at 16 04 19

**Android phone -> "phone"** 
Screenshot_20191118-155909_flu@home
Screenshot_20191118-155602_flu@home

**Android tablet -> "tablet"** 
Screenshot_20191119-105804
Screenshot_20191119-105836

Please feel free to comment if you feel we should leave "device" in some contexts.


tpaik18:
Oh btw, @smeds1 , this will break any code (Metabase?) that is dependent on `device.idiomText` in the JSON. Instead of `handset`, `tablet` or `unknown`, it will return an ugly string like `ios++handset` or `android++tablet`. Concerns?

smeds1:
@tpaik18 I've never used idiomText. You should be safe.

PR #408, opened 2019-11-19
FEV-1445 Update design for country picker

audererob:
Changing the design of the country picker to a text field rather than a multi-select option list. I'll remove support for the multi-option component in a follow-up PR.

Also fixing the broken quad-tap-to-reset bug that was introduced with the reset state refactoring.


tpaik18:
Can't tell what the keyboard would be for the country input field, but it would be nice if it capitalized the first letter of each word.

PR #413, opened 2019-11-19
FEV-1482 "Connect to verify barcode" screen is missing copy

audererob:
[FEV-1482](https://auderenow.atlassian.net/browse/FEV-1482)
Adding missing description and alert for barcode connection error screen.

Also taking the opportunity to remove the client-side demo mode barcode support now that the server supports it.


tpaik18:
@smeds1 I think I found the table in the local DB that needs to be populated with rows in order for automation to work. I'll send it to you in a DM.

FluStudy_us/src/util/patientAchievementInfo.ts
```diff
@@ -115,7 +117,9 @@ export async function getEmailConfirmationNextScreen() {
 export async function getShippingTextVariables() {
   const state: StoreState = (await getStore()).getState();
   return {
-    state: state.meta.patientAchievementInfo.state.toLowerCase(),
+    state: state.meta.patientAchievementInfo.state
+      .toLowerCase()
+      .replace(" ", "-"),
```
tpaik18: I think we can omit the replace because @billy-audere already validates when reading the G Sheet that `state` is two letters, so I don't think it can have a space in it.

FluStudy_us/src/util/patientAchievementInfo.ts
```diff
@@ -125,7 +129,9 @@ export async function getShippingTextVariables() {
 export async function getThankYouTextVariables() {
   const state: StoreState = (await getStore()).getState();
   return {
-    state: state.meta.patientAchievementInfo.state.toLowerCase(),
+    state: state.meta.patientAchievementInfo.state
+      .toLowerCase()
+      .replace(" ", "+"),
```
tpaik18: Same as above.

smeds1:
@audererob I'm fine with removing the demo barcode. It seems like there are ways to get the gsheet info into my local db.

PR #415, opened 2019-11-20
FEV-1447 State name is not appearing in firebase for travel outside home state question

audererob:
[FEV-1447](https://auderenow.atlassian.net/browse/FEV-1447)
When a question's answer is updated, the question's title/description are saved into the survey. We now include any associated text variables and use them in the i18n.t() call to correctly format the text. The screen's questions component now owns the text variables and passes them along to each question type component.


audererob:
SurveyQuestions are immutable, and it would be good to keep them that way - I think the variables could be made part of SurveyResponse though, and perhaps be initialized in the Questions component. I don't want to introduce overhead for unanswered questions so it might be tricky to pull off. I'll give it a shot.

FluStudy_us/src/store/questions.ts
```diff
@@ -104,3 +120,37 @@ export function updateAnswer(
...
+    ).trim(),
+  };
+}
+
+export function setResponseTextVariables(
+  question: SurveyQuestion,
+  textVariables: any
+) {
+  return {
+    type: "SET_RESPONSE_TEXTVARS",
```
tpaik18: Nit: Better to name this `setQuestionTextVariables` and `SET_QUESTION_TEXTVARS`?

audererob: It's the SurveyResponse object that's being updated here, so I think this naming is correct, though a little confusing. It's because the question's text is persisted with the response.

PR #423, opened 2019-11-21
Adding keep-alive behavior to endpoints which update derived tables.

billy-audere:
Updating reporting schema may run long, often the server process succeeds but the client believes the process failed after connection is closed.  This should clear up some noisy infrastructure alarms.


billy-audere:
Ultimately moving these processes out of FluApi & Express may be desirable.

PR #425, opened 2019-11-21
FEV-1416 Date picker saves wrong date in timezones east of UTC

audererob:
[FEV-1416](https://auderenow.atlassian.net/browse/FEV-1416)
The DatePicker uses local date, but we want to store a UTC date with the local date in the database - thus we'll manually construct it.

Also fixes [FEV-1458](https://auderenow.atlassian.net/browse/FEV-1458) User can pick 6/30/19 for flu shot date on iOS


FluStudy_us/src/ui/components/DatePicker.tsx
```diff
@@ -35,23 +35,43 @@ class DatePicker extends React.Component<Props & WithNamespaces> {
...
     return (
       <View style={[styles.container, !!highlighted && HIGHLIGHT_STYLE]}>
         <DatePickerComponent.default
-          date={date}
+          date={selectedDate}
           mode="date"
           placeholder={t("selectDate")}
           confirmBtnText={t("common:button:done")}
           cancelBtnText={t("common:button:cancel")}
           format="YYYY-MM-DD"
```
tpaik18: By the way the `format` here is where I think we can put in MMM D, YYYY or some such if we wanted to.

PR #426, opened 2019-11-21
Update RDT filter

jennya:
* Don't require the rdt outline to be in the expected location for the high resolution frame (fixes https://auderenow.atlassian.net/browse/FEV-1456)
* Save rdt test area bitmap as a jpeg and notify RN by reusing the resultWindowImageUri variable
* Run imageFilter on the entire image for preview frames where the RDT is not found
* Only run the sharpness check (not the brightness check) on the still frames
* Reduce high res frame sharpness threshold


FluStudy_us/android/app/src/main/java/host/exp/exponent/DetectorView.java
```diff
@@ -436,15 +434,15 @@ protected abstract void processResult(
             return mappedRecognitions;
         }
 
-        protected String saveImage() {
-            File photo = new File(activity.getFilesDir(), RDT_PHOTO_FILE_NAME);
+        protected String saveImage(Bitmap bitmap, String filename) {
+            File photo = new File(activity.getFilesDir(), filename);
```
rightparen: Consider taking a filename and providing an overload that does the default one.

FluStudy_us/android/app/src/main/java/host/exp/exponent/DetectorView.java
```diff
@@ -461,25 +459,29 @@ protected void initialize() {
...
-                filterResult = imageFilter.validateImage(rdtResult.rdtStrip, false);
-                if (!stillCaptureInProgress && filterResult.isSharp() && filterResult.exposureResult.equals(ImageFilter.ExposureResult.NORMAL)) {
-                    Log.d(TAG, "Have good preview frame, making single request");
-                    stillCaptureInProgress = true;
-                    cameraController.captureStill();
-                }
+        protected void processResult(IprdAdapter.FrameResult iprdResult, List<Classifier.Recognition> mappedRecognitions) {
+            RDTTracker.RDTPreviewResult rdtResult = rdtTracker.extractRDTFromPreview(mappedRecognitions, imageBitmap);
+
+            if (rdtResult == null || rdtResult.rdtOutline == null || rdtResult.rdtStrip == null) {
```
rightparen: There should still be an IPRD check here and in the other processResult.  Fwiw, I'm changing it consistently to:
```
if (iprdResult == null || iprdResult.isAccepted) {
```

jennya: I think we're covered because we only call processResult if that condition is true

FluStudy_us/android/app/src/main/java/host/exp/exponent/DetectorView.java
jennya: Will defer to your new implementation of saveImage

PR #427, opened 2019-11-22
FEV-1495 Copy changes from UI review

tpaik18:
Copy tweaks from the UI review meeting 11/21. 
Also deleting unused unverifiedBarcodeAlert() and the now unused inputManually().


FluStudy_us/src/util/barcodeVerification.ts
```diff
@@ -32,19 +32,3 @@ export function invalidBarcodeShapeAlert(
...
-    i18n.t("barcode:notVerified"),
-    i18n.t("barcode:pleaseRetry", { action }),
-    [
-      {
-        text: i18n.t("barcode:tryAgain"),
-        onPress,
-      },
-    ]
-  );
-}
```
smeds1: I'm not sure I understand what you're taking out here. Was this some sort of dead code, or is this the alert that pops up if you give the barcode in the wrong format?

smeds1:
Oh, I just saw your comments in the PR. Yes, let's take out dead barcode code.

FluStudy_us/src/util/barcodeVerification.ts
tpaik18: Yes this was a dead function that was not being called by anyone.

PR #428, opened 2019-11-22
FEV-1493 Design tweaks from 11/21 UI review

audererob:
[FEV-1493](https://auderenow.atlassian.net/browse/FEV-1493)
Made the following changes:
* Research study page is on purple background
* Remove divider from survey pages that don't have hint text or description
* Fixed vertical space on "What do you see?" page
* Bolded questions on last page
* Changed font size back to cough dimensions
* Tweaked spacing of a few things to be consistent
* Updated "Prepare kit return" asset with latest
* Left-aligning all title text
* Change link color on purple pages
* Update date format for date picker


FluStudy_us/src/resources/MenuConfig.ts
```diff
@@ -102,14 +102,5 @@ export const MenuScreens: ScreenConfig[] = [
...
-  menuScreen("Version", false, [
-    {
-      tag: ScreenText,
-      props: {
-        label: "description",
-        center: true,
-      },
-    },
-    { tag: BuildInfo },
-  ]),
```
tpaik18: Not sure I understand this change. This seems like it would take out the "Copyright c 2019 Audere. All rights reserved." text. Did we want to remove that?

audererob: I changed the "hasDesc" parameter to true, which will add in the description text automatically. The text was only specified here in order to center it, but now that the title is left-aligned this text doesn't need to be. The About page looks as it did before except the title and copyright are left-aligned.

PR #438, opened 2019-11-23
Integrate IPRD RDT reader filter code and tflite model

rightparen:


rightparen:
@jennya here is the specific comment from IPRD, and the whole thread provides more context:  https://iprdsolutions.slack.com/archives/CHSQ0A1MH/p1574438042042200?thread_ts=1574415221.041400&cid=CHSQ0A1MH

PR #443, opened 2019-11-25
FEV-1488 Barcode manual entry should squeeze out any spaces

audererob:
[FEV-1488](https://auderenow.atlassian.net/browse/FEV-1488)
This change prevents non-alphanumeric characters from being entered in barcodes.


audererob:
Yes, the i at the end of the regex means case insensitive.

tpaik18:
Ah.

PR #452, opened 2019-11-26
Update filters

jennya:
Tidy up the filter code and update for processing filters:

```
If iprd.isSteady()
  If audere.rdtFound()
    If audere.rdtInDesiredLocation()
      If iprd.isAccepted() → take high resolution still
      Else if iprd.noRdt() [degenerate] → “Hold steady” ?
      Else if !iprd.isSharp() → “Hold steady”
      Else if !iprd.normalExposure() → “Adjust lighting”
    Else if audere.tooSmall() → “Raise your device”
    Else if audere.tooBig() → “Lower your device”
(all) Else → “Center device over the test strip and hold steady”
```

And added a debug string next to the FPS to display why a preview frame was rejected


FluStudy_us/android/app/src/main/java/host/exp/exponent/IprdAdapter.java
```diff
@@ -5,72 +5,183 @@
...
   public static final String TAG = "IprdAdapter";
+  private static final String IPRD_MODEL_FILE = "iprd.tflite";
+
+  private Activity activity;
+  private BaseLoaderCallback loaderCallback;
+  private boolean haveOpenCv = false;
+  private RdtAPI iprdApi;
+
+  public IprdAdapter(Activity activity) {
+    this.activity = activity;
```
rightparen: Discussed in person, note for posterity:
* Prefer to take a `Context` and save `context.getApplicationContext()`.  Even though our app only has one Activity, it's a better pattern because it avoids memory leaks.
* The `IprdAdapter` class was intended to be a thin abstraction layer over the IPRD code.  Consider factoring out this loader code.

jennya: Ok moved to ResourceLoader

PR #453, opened 2019-11-26
Update schema for shipped_kits and chills_derived.surveys

smeds1:
@billy-audere Yesterday we agreed that we would like to import all the PII data into our data tables and restrict access with permissioning. I took a stab at updating the schema so that those other fields in the shipping report would get imported into our database.


FluApi/src/models/db/chills.ts
```diff
@@ -295,11 +295,11 @@ export function defineShippedKits(sql: SplitSql) {
     {
       evidationId: stringColumn("evidation_id"),
       barcode: stringColumn(),
-      email: stringColumn(),
-      birthdate: stringColumn(),
-      sex: stringColumn(),
-      city: stringColumn(),
-      state: stringColumn(),
+      email: stringColumn("email"),
```
billy-audere: These don't need to be specified if the column name matches the property name.  That said there is also no harm in doing so.

billy-audere:
@smeds1 Let me know if you'd like to go over how to deploy these changes.

PR #458, opened 2019-11-27
Fix permission listener, only request storage permission in debug builds

jennya:


jennya:
@rightparen demoMode was a little tricky because of initialization challenges between RN and Android. Are you ok with having the storage permission requested for debug builds only?

FluStudy_us/android/app/src/main/java/host/exp/exponent/DetectorView.java
```diff
@@ -92,8 +94,9 @@
 
     public DetectorView(Context context, AttributeSet attrs) {
         super(context, attrs);
-        if (context instanceof Activity) {
-            activity = (Activity) context;
+        if (context instanceof MainActivity) {
+            activity = (MainActivity) context;
+            activity.addPermissionListener(this);
```
rightparen: `else` case now has an obsolete error message.

PR #467, opened 2019-12-03
FEV-1516 tips are too fast

tpaik18:
Isn't a great fix, but tried to shave a few words off the tips without changing the meaning; LMK if you don't like the wording please. Also bumped the time to 11 seconds each so we don't have the tip change exactly when the timer is up. 

Delete unused FirstTimer.desc.


auderephilip:
<a href="https://gitme.me/image?url=https%3A%2F%2Fmedia2.giphy.com%2Fmedia%2F3oEdva9BUHPIs2SkGk%2Fgiphy.gif&token=thanks" data-gitmeme-token="thanks"></a>

PR #469, opened 2019-12-03
Read Virena files from S3 and load them into the database.

billy-audere:
Supports binary XLSB for backfill and XLSX files. Writes files to the database in chunks and compares ETag/hash + key to determine which files need to be loaded.


billy-audere:
- Breaking files into chunks is a little confusing or complex to track - however some of the backfill files are a big larger ~1 million rows and ran into issues locally completing (my local process would hang).  As well it would make it difficult to keep the connection alive.
- `bulkCreate` uses `insert into ... values (...), (...), ...` which is shorthand for multiple insert statements. A more efficient solution would be to use `copy into ...`.  The code in this PR is probably somewhat dead in that we may want to use CSV copy import for this use case in the future.  With our container I didn't want to write files to the local filesystem (some of the backfill XLSB files expand out to a few hundred MB).  There is an AWS extension for RDS that can import files from S3 but only supports Postgres engine 11.1+.  In the future I would probably recommend upgrading the Postgres version on the RDS instance and giving the AWS extension a try.
- Files will be emailed weekly and need to be manually copied to S3.  The weekly files over the past month have been around 15000-20000 rows.

FluApi/db/migrations/nonpii/20191121112031-virena-import.js
```diff
@@ -0,0 +1,85 @@
...
+          references: {
+            model: {
+              tableName: "virena_files",
+              schema: schema,
+            },
+            key: "id",
+          },
+          onDelete: "CASCADE"
+        },
+        segment: column(Sequelize.INTEGER)
```
rightparen: This is probably okay for now, but is a bit brittle.  E.g. if AWS constraints ever forced us to change segment size, this table becomes confusing.  The segment code is parameterized by segment size, but if that value were not actually a universal constant, we could silently import the wrong data.

 if we do something like this in the future, consider having a nullable number on "virena_files" that tracks the next row number we need to import.  This would start at 0, update we import a segment of whatever size we currently support, and if the update hits the end of the file becomes NULL.  Then it becomes trivial to tune segment size.

FluApi/src/endpoints/chillsVirenaApi.ts
```diff
@@ -0,0 +1,50 @@
...
+  constructor(sql: SplitSql) {
+    const secrets = new SecretConfig(sql);
+    this.service = new LazyAsync(async () => {
+      const s3Config = await getS3Config(secrets);
+      const s3 = new AWS.S3({ region: "us-west-2" });
+      const client = new VirenaClient(s3, s3Config);
+      return new ChillsVirenaService(sql, client, this.SEGMENT_SIZE);
+    });
+
+    this.pipeline = new CoughDataPipeline(sql.nonPii);
```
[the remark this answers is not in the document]
billy-audere: Apologies for missing this.  Sam and I discussed whether there were any aggregations we should perform on the Virena data.  For now we're just going to copy it over directly.  I have split out copying Virena into a separate pipeline and updated.

FluApi/db/migrations/nonpii/20191121112031-virena-import.js
billy-audere: Makes sense.  I have updated the PR to use a `nextRow` column in the files table.

PR #471, opened 2019-12-03
Mirroring RDT info updates to the Chills protocol

billy-audere:


smeds1:
There are a couple other fields we are no longer using this this file:
clientversion_rdt_android, clientversion_rdt_ios, rdtreaderresult_skippedduetomemwarning

Can we take those out as well?

billy-audere:
@smeds1 I have removed the client version fields.  For the mem warning field it looks like it can still be set [here](https://github.com/AudereNow/audere/blob/143560d4c5fdf026948797d93d2bc5732858da75/FluStudy_us/src/ui/components/flu/AndroidRDTReader.tsx#L357).  If this should also be removed let's follow up separately, along with any other fields you find missing.

PR #475, opened 2019-12-03
Bumping Metabase version

billy-audere:
Appears to be the most recent version of Metabase available.

To deploy requires applying Terraform changes.  I don't believe updating the ECS task definition will force running tasks to be recreated so there is probably a Terraform taint at the least that also needs to happen as part of deployment.


billy-audere:
Looking at the logs when deploying it seems like updating the service does in fact attempt to recreate the task.  However, in the specific case of Metabase there is no enough memory to reserve on the instance and the task creation fails.

PR #481, opened 2019-12-04
Debug FEV-1538

tpaik18:
Echoing the caught exception to the screen to try to see what is wrong with FEV-1538.


FluStudy_us/src/ui/components/flu/TestStripCamera.tsx
```diff
@@ -187,7 +187,7 @@ class TestStripCamera extends React.Component<Props & WithNamespaces> {
         this.setState({ spinner: false });
         navigation.dispatch(StackActions.push({ routeName: next }));
       } catch (e) {
-        Alert.alert("", t("error"), [
+        Alert.alert("", t("error") + e, [
```
audererob: Do you need a CR or two prepended to avoid the first line of the exception text being appended to the last line of the error string?

PR #485, opened 2019-12-05
On photo capture error, loading icon dismissed on alert dismissal

goldjay:
[FEV-1538](https://auderenow.atlassian.net/browse/FEV-1538)

Still not sure why photo capture fails the first time when the phone is sitting on a desk, but now we've got appropriate error handling.

* NOTE: VSCode Extension that organizes imports


tpaik18:
Hoping someone more technical than me can review.

rightparen:
@goldjay were you able to verify that you could repro before this change and not after?  I don't understand why the previous code caused us to hang.

My hypothesis on failure is that there is nothing the phone can focus on, so focusing/metering would fail.  Just a guess though.

PR #486, opened 2019-12-05
Add S3 bucket for UW to upload photos

rightparen:
Not yet tested.  I want to get feedback before polluting AWS state.


rightparen:
Actually, it's pretty easy to clean this up, so I just applied and it seems okay.

PR #497, opened 2019-12-09
FEV-1536: move box down on barcode scanner

smeds1:
This PR moves the help box on the barcode scanner so that scanning is easier on iOS devices. I only tested it on my iPhone 8+. We'll need to do some more testing in a build to make sure that there is no regression on other iOS devices or Android.


smeds1:
@tpaik18 I adjusted the margins so that the text appears properly on the SE and 8+. I'm not very confident that it will look right on our whole boatload of devices, though.

PR #500, opened 2019-12-09
Save extracted test area to Firebase

jennya:
Upload the extracted and rotated image of the test area for debugging interpretation

### DIFF
--- a/FluStudy_us/src/ui/components/flu/AndroidRDTReader.tsx
+++ b/FluStudy_us/src/ui/components/flu/AndroidRDTReader.tsx
@@ -507,7 +507,7 @@ class AndroidRDTReader extends React.Component<Props & WithNamespaces, State> {
     dispatch(setRDTCaptureTime(true));
     try {
       const photoId = await newUID();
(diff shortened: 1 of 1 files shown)


FluStudy_us/src/ui/components/flu/AndroidRDTReader.tsx
```diff
@@ -507,15 +507,23 @@ class AndroidRDTReader extends React.Component<Props & WithNamespaces, State> {
     dispatch(setRDTCaptureTime(true));
     try {
       const photoId = await newUID();
+      const resultPhotoId = await newUID();
```
rightparen: `testAreaPhotoId`?

rightparen:
At some point I would also like to save all the dimensions and matrices.

FluStudy_us/src/ui/components/flu/AndroidRDTReader.tsx
jennya: done

PR #501, opened 2019-12-09
Only draw interpretation results on test area bitmap when in debug mode

jennya:
I realized that we may want to be able to go back and run new phase 2 models on our production images, and so should only do the debug interpretation drawing when in debug mode. Will log the full results from the interpretation phase in a follow up pr so that we could reconstruct this if desired offline

### DIFF
--- a/FluStudy_us/android/app/src/main/java/host/exp/exponent/tracking/InterpretationTracker.java
+++ b/FluStudy_us/android/app/src/main/java/host/exp/exponent/tracking/InterpretationTracker.java
@@ -5,7 +5,6 @@
 
 package host.exp.exponent.tracking;
 
-import android.graphics.Bitmap;
 import android.graphics.Canvas;
 import android.graphics.Color;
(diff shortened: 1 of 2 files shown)


FluStudy_us/android/app/src/main/java/host/exp/exponent/tracking/InterpretationTracker.java
```diff
@@ -87,7 +88,10 @@ private static Paint getPaint() {
         return p;
     }
 
-    private static void drawLine(Classifier.Recognition recognition, int color) {
+    private static void drawLine(Classifier.Recognition recognition, int color, boolean drawResults) {
```
rightparen: Maybe `drawLineIf(boolean condition, Classifier.Recgnition...`?

jennya: Done

PR #514, opened 2019-12-12
Fix title overlap for RDT image display

jennya:


tpaik18:
@smeds1 We would like this for release, right? @jennya Could you please cherry-pick into `chills-release`?

jennya:
@tpaik18 done

PR #521, opened 2019-12-13
Update phase 2 extraction to match current python version

jennya:
I should have checked the test area locations in flu-rdt/feature-map.json when I updated the strip width in https://github.com/AudereNow/audere/pull/510. With this update, all values from the feature-map match what we have in the Android code.

The extractor script grabs more of the test area which seems useful in case there is any amount of vertical offset from center. Regardless, the device extraction logic should match whatever was done to the training data.


jennya:
Oh dear, I just realized this isn't what's in the feature-map.json file that's currently checked in to openrdt-internal, just what I have locally... Sorry for being sketchy. These are the values I used when I ran extraction and trained the phase 2 model that's currently live in the app, so I think we should use them. I will create a task to update this code to read the constants from a feature-map.json file or equivalent that should be added as a resource to the app along with the tflite and label files

PR #537, opened 2019-12-27
Check whether FluApi really deployed

tpaik18:
Step 1 for https://auderenow.atlassian.net/browse/FEV-1489 

Implementing logic that @rightparen suggested: 

> Check task ID before deploy, and wait for it to change after deploy, and make sure status of the new guy is RUNNING. 
> Put this in a loop to periodically check/sleep. if it takes more than 2 minutes, then fail.


tpaik18:
Closing this in favor of another that includes the post-deploy curl test framework

PR #540, opened 2019-12-31
FEV-1489 Check whether FluApi deploy was successful

tpaik18:
Modify `manage-service` to do the following as suggested by @rightparen : 

> Check task ID before deploy, and wait for it to change after deploy, and make sure status of the new guy is RUNNING. 
> Put this in a loop to periodically check/sleep. if it takes more than 2 minutes, then fail. 
> Add a new blob into service-registry.json under “fluapi” which is called test, and it’s the name of the test script to run. The filename in the audere/scripts/folder. 
> Add curl lines into there. Take as an argument staging or prod. 
> Then in the ecs_deploy check for existence of test and run that after confirming deploy. 

Couple notes: 
1. `test-fluapi-post-deploy` script reads the actual test values from `fluapi-curl-tests.json`.  Wasn't sure of the best framework for this. Currently the json allows for specifying separate expected value for staging vs prod, which I expect would be needed for Cough giftcard API and Chills barcode API calls. 
2. Pretty sure @rightparen will flag `test-fluapi-post-deploy` line 31 for neglecting to quote `$expected_result`, but this is intentional and allows for us to use the bash * syntax in the expected value  (starts with or ends with). Acceptable? Or hacky? 
3. The tests in `fluapi-curl-tests.json` are just for example. Hoping @billy-audere @balderdash and @rightparen can flesh these out with actual API tests.


scripts/manage-service
```diff
@@ -71,9 +74,40 @@ function ecs_deploy() {
...
+      (( count++ )) 
+    fi
+  done
+
+  if [[ $count -gt $MAX_RETRIES ]]; then 
+    fail "${RED}ERROR:${NC} New task not deployed after $(($RETRY_INTERVAL * $MAX_RETRIES)) seconds"
+  else 
+    test_script="$(config_get "\"$service\".\"test\"")"
+    if [[ ! -z "$test_script" ]]; then 
+      echo "Found post-deploy test script $test_script"
```
rightparen: `"Running post-deploy..."`

If the script is requested but doesn't exist, the error message would be confusing otherwise.

scripts/test-fluapi-post-deploy
```diff
@@ -0,0 +1,52 @@
...
+
+function fail() {
+  echo -e "$*"
+  exit 1
+}
+
+function do_tests() {
+  local target_type="$1"
+  local BASEURL="$2"
+  local TEST_FILE_JSON="$SELF_DIR/fluapi-curl-tests.json"
```
rightparen: This is fine but may be overkill.  Since this whole file is fluapi specific this could potentially be more hard-coded.

tpaik18: Yes, now that I imagine the actual API calls that might be needed for tests (including doc IDs, barcodes, and secrets) this json framework may be more limiting than helpful. @balderdash and @billy-audere, your thoughts?

billy-audere: The format for the JSON assertions may be limiting.  I think it would be reasonable to make assertions on the HTTP status code, or other complex criteria, which may be easier to express in bash (or TypeScript, etc).

Calling out to other processes or modifying this to execute a hand coded script would be relatively easy.  I think the facility is here for whatever we would like to do & in the short term the JSON format may suffice and be easiest to modify.

At some point executing tests with Jest makes the most sense to me but this implementation seems like a wonderful starting point.  I wouldn't recommend any changes inside of this PR & rather defer that until we implement new tests and have more clarity about what is needed.

tpaik18: Billy called my PR "wonderful". I'm merging!

PR #547, opened 2020-01-09
Running shipped kits import for Chills every day

billy-audere:
Also updating the labels for cron schedules. Previously the schedules were mislabeled and used PST which fluxuates due to daylight savings.


terraform/modules/flu-lambda/main.tf
```diff
@@ -8,13 +8,9 @@ locals {
...
   slack_archive_path = "../../../local/lambda/cloudwatch-slack.zip"
 
-  // This is 8:30 AM and 1:30 PM local in PST
   // See: https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/ScheduledEvents.html
-  cron_weekdays_before_9AM_and_1PM_PST = "cron(30 15,19 ? * MON-FRI *)"
-  cron_weekdays_before_1PM_PST = "cron(30 19 ? * MON-FRI *)"
-  cron_weekdays_at_4AM_PST = "cron(0 10 ? * MON-FRI *)"
-  cron_everyday_at_5AM_PST = "cron(0 12 ? * * *)"
-  cron_monday_thursday_before_9AM_PST = "cron(30 15 ? * MON,THU *)"
+  cron_everyday_at_10PM_UTC = "cron(0 10 ? * * *)"
```
tpaik18: I think you mean 10AM?

PR #549, opened 2020-01-09
Catch camera already closed exceptions

jennya:
For https://auderenow.atlassian.net/browse/FEV-1552 -- we have multiple crash logs of attempts to close the camera when it's already closed. We should swallow that exception and continue


audererob:
For the close case it doesn't seem bad to catch everything since we're tearing it all down anyway. I agree that for the other cases we risk missing some legit bugs by catching everything.

jennya:
Makes sense -- I'll switch to catching the IllegalState explicitly

PR #551, opened 2020-01-09
Allowing users to specify a different AWS user for MFA authentication…

billy-audere:
… than the local user name


smeds1:
@rightparen This fixed all the authentication problems I was having when you tried to get me to run the flu@home website deploy script that you wrote.

PR #553, opened 2020-01-15
Removing unused columns in MTL report

billy-audere:
These columns were specified in example files but are not appearing populated in production files or are not considered useful. Removing them to simplify the format for MTL reports.


billy-audere:
Should also correct sourcing of strip temperature status.

PR #554, opened 2020-01-15
Drop derived objects that depend on mtl_reports table

billy-audere:
Views that depend on a table and expose a column prevent that column from being deleted in the original table. This causes the migration to fail.


billy-audere:
@tpaik18 I believe `--no-progress-meter` is a newer flag in curl.  I always run into complaints when deploying so I have changed it to `-sS`.

PR #560, opened 2020-01-20
Enable end-to-end encryption of test result objects (1 of ~3)

aaronkroth:
Enable encryption of selected parts of test result objects. This first part outlines the
API: clients of the encrypt module are expected to add an
EncryptionPolicy.MustEncrypt field to objects within a record they
wish to remain encrypted everywhere between client and final database.
So, for instance:

 ```
testResults = {
   "date": ...,
   "hivStatus": {
      "encryptionPolicy": EncryptionPolicy.MustEncrypt,
      "positive": false,
      ...
    },
  }
```

Then later:

  `applyEncryptionPolicy(testResults)
`
Only the objects marked for encryption are encrypted. This is to
facilitate intermediate processing on non-encrypted nodes.

Part 2 of this diff series will plug in a real encryption
algorithm. (For now we just "encrypt" by converting strings to arrays
of UTF-8 bytes.)

Part 3 (and higher?) of the series will implement key management:
generation of keys within database and fetching thereof on clients.

### CODE REVIEW

File: lib/DataProtection/src/encrypt.ts [hunk 1]
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  function visit(obj: any): void {
+    const policy = obj.encryptionPolicy || EncryptionPolicy.None;
+    switch (policy) {
+      case EncryptionPolicy.None: {
+        for (let k in obj) {
+          const child = obj[k];
+          if (child instanceof Object && !(child instanceof Array)) {
+            visit(child);
```
rightparen: I don't think we'll hit an issue here, but recursion could be bad if we ever added linked lists.  Similar concern below in unapply..

lib/DataProtection/__tests__/encrypt.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    }
+  };
+  applyEncryptionPolicy(obj["subobject1"]["subobject2"]);
+  unapplyEncryptionPolicy(obj);
+  expect(obj).toEqual(original);
+  applyEncryptionPolicy(obj["subobject1"]["subobject2"]);
+  applyEncryptionPolicy(obj);
+  unapplyEncryptionPolicy(obj);
+  expect(obj).toEqual(original);
+})
```
rightparen: These tests were very helpful for me to get a quick understanding of what your code is doing.  Thanks!

lib/DataProtection/src/encrypt.ts
```diff
@@ -0,0 +1,178 @@
...
+
+function decryptObject(obj: any): void {
+  const key = getKey(obj.keyId);
+  const decryptedObj = JSON.parse(new TextDecoder().decode(Uint8Array.from(obj.payload)));
+  emptyObject(obj);
+  Object.assign(obj, decryptedObj);
+}
+
+// Should be used exclusively for testing.
+export const Internals = {
```
rightparen: To borrow from Android, consider "VisibleForTesting".

lib/DataProtection/src/encrypt.ts
```diff
@@ -0,0 +1,178 @@
...
+  }
+}
+
+export interface Key {
+  id: string;
+  isExpired(): boolean;
+}
+
+export const CURRENT_KEY: unique symbol = Symbol("Current Key");
+export const NEW_KEY: unique symbol = Symbol("New Key");
```
rightparen: TIL.  Curious to see where this is going in later diffs.

lib/DataProtection/__tests__/encrypt.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+  expect(obj).toEqual({})
+})
+
+test('After ensureCurrentKey() called, current key should not be expired (else exception was thrown).', () => {
+  try {
+    ensureCurrentKey();
+    const key = Internals.getKey(CURRENT_KEY);
+    expect(key.isExpired()).toBe(false);
+  } catch (exc) {
+    // TODO: check something like type of exc?
```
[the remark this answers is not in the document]
aaronkroth: Well, not really valid for current code under test; this is just a placeholder for the third envisioned diff where we possibly make a call to network or local storage in order to retrieve current key.

lib/DataProtection/src/encrypt.ts
aaronkroth: Not sure I understand. You mean, if there are recursive references? But won't that screw up `JSON.stringify()` in the first place? I think I don't understand.

lib/DataProtection/__tests__/encrypt.spec.ts
aaronkroth: Yeah, programming in such a free-wheeling language (from the point of view of someone whose language-ideal is Haskell) makes me awfully nervous. Tests allay the nervousness a little. :)

lib/DataProtection/src/encrypt.ts
aaronkroth: Np.

lib/DataProtection/src/encrypt.ts
aaronkroth: Probably going away -- or at least to be heavily modified since I discovered `hybrid-crypto-js`. :) Working on this now.

aaronkroth:
Re the various functions taking in objects by reference and mutating them in place: totally share your concern with mutable state. Can certainly rewrite to return new objects and assign, but that requires having handle on parent node in order to have reference rather than copy of object to be replaces. As far as I can see, this would get you a little less mutability at the cost of substantially complicating the management of the recursive calls. There's also the somewhat minor argument in favor of in-place mutation that it "guarantees" that once the user has passed their object into the encryption function, there won't be any plaintext versions lying around afterward. Of course, there are many ways around this, and it's not much of an argument, but there you are.

aaronkroth:
Oh, and thanks for the thorough review! This is my first Audere code, and in indeed my first TypeScript code of any kind, so.

lib/DataProtection/__tests__/encrypt.spec.ts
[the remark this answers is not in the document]
rightparen: For this test code though, would we ever run against a network that could fail?  I would have expected that we would always configure the test code to run against a mock or local store, and that we would want to report if the mock/local ever failed by allowing the exception to propagate and fail the test.

lib/DataProtection/src/encrypt.ts
rightparen: Oh you're right, never mind.  I should have expected that JSON.stringify uses simple recursion as well.

rightparen:
Wouldn't the parent reference thing only be an issue if you were mutating at a higher level?  I was imagining each level of the recursion just returns a potentially different version of its `root`.

Code about to put something in Firestore would encrypt just before storing, so I'm not worried about unencrypted data escaping.

The total complexity might even be lower if it lets us omit some of the own-property and object-clearing code here.  Even if we don't save any of that though, I think the immutability is worth it.

lib/DataProtection/__tests__/encrypt.spec.ts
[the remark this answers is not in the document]
aaronkroth: Yeah, was eventually planning to mock out exceptional and non-exception cases. But can cut this from code till I get to that point...

aaronkroth:
This gets rid of the internal mutation of objects (and does, as you predicted, make the code simpler -- thanks!). Not sure about deployment strategy: I'd almost consider putting this in its own top-level directory in case it starts to use heavier-weight libraries related to crypto and networking that there's no reason to bring in if people just want a little bit of lib. What do you recommend? Thanks again!

lib/DataProtection/src/encrypt.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  function visit(obj: any): any {
+    const policy = obj.encryptionPolicy || EncryptionPolicy.None;
+    switch (policy) {
+      case EncryptionPolicy.None: {
+        for (let k in obj) {
+          const child = obj[k];
+          if (child instanceof Object && !(child instanceof Array)) {
+            obj[k] = visit(child);
```
rightparen: Hmm, this still mutates the argument.  Consider initializing a new object at the top:
```
const result = { ...obj };
```
and then set `result[k]` here and then `return result;`.

Similar comment for unapply.

aaronkroth: Ah, sorry, I misunderstood your original point about mutation. Will change as you suggest.

aaronkroth:
Encryption and decryption should now create new objects with no shared references to input objects (but see comment `Note this is the one place where...`).

aaronkroth:
If I'm reading [this](https://www.npmjs.com/package/hybrid-crypto-js) right, it looks like `hybrid-crypto-js` works on both Node and Reactive Native? That would be good news, as this seems like exactly the package I need, and it's already used elsewhere in the Audere codebase...

> Re: deployment and crypto libraries, wow, good issue. The client and server will need to use different `crypto` objects. The server can use the `crypto` built into NodeJS, but the client library will need a crypto library that runs in the funky React Native environment. I say funky because it's not NodeJS and it's not browser, and unfortunately crypto is where they most diverge. I suspect anything higher level that depends on `crypto` will diverge as well.
> 
> So it might be best for now to leave this in `/lib` and have calling code pass in `crypto` plus any other deps as parameters rather than having code here `import` or `require`. I do like the idea of separating the crypto protocol from app protocol. That said, I think that can happen later, and I'd like others to weigh in before we start moving around top-level directories in the repo.

lib/DataProtection/src/encrypt.ts
```diff
@@ -0,0 +1,200 @@
...
+    } else if (obj instanceof Array) {
+      return obj.map(e => visit(e));
+    }
+
+    const policy = obj.encryptionPolicy || EncryptionPolicy.None;
+    switch (policy) {
+      case EncryptionPolicy.None: {
+        // Go to a little trouble here to make full deep copies of
+        // fields, including Arrays.
+        let copy = {};
```
rightparen: Oops, this doesn't actually work for Array.  An object `{}` with integer keys is different than an Array you would get from deserializing JSON.

This is where I start worrying that we are building something more general than we will use--we will generally have one or two top-level properties that are encrypted.  But the general case here would need to distinguish Array and initialize as `[]` rather than `{}`.

aaronkroth: Forgive my Javascript ignorance, but would the `obj instanceof Array` test just before this code not cover this case?

rightparen: Ah, oops!  My bad, you're right.

For `unapply()` below though it looks like the Array check only happens at top-level, so Arrays inside subobjects there might become Object.

aaronkroth: Pretty sure the code as written, though a little unintuitive, does handle array at all levels correctly. I any case, I've just pushed an additional test for additional peace of mind.

By the way, possibly one advantage of allowing encryption à la carte below the top-level is that it could form the kernel of data protection policy. You could imagine adding additional policies like `Delete` or `DeleteField(field)` that tell you how to sanitize data for hand-off downstream. Not convinced this is the right way to do it, but thought I'd throw it out there.

PR #562, opened 2020-01-21
FEV-1610: Add progress function to indicate request is still alive

smeds1:
The purpose of this PR is to eliminate the CloudWatch alarm that is triggered by the CDC data lambda import. The issue is likely that the request takes awhile, and the server needs reassurance that the request is still alive. 

@billy-audere I tried to add the progress function that you were talking about, but I don't think it is 100% correct yet. It probably does not need to be run after every single row of data is upserted, right?


FluApi/src/endpoints/chillsSurveillanceApi.ts
```diff
@@ -28,16 +29,18 @@ export class ChillsSurveillanceEndpoint {
...
+
     const svc = await this.service.get();
     logger.info("Finding CDC reports to import");
-    await svc.import();
+    await svc.import(progress);
 
-    const pipelineSvc = new DataPipelineService();
     logger.info("Refreshing derived schema");
-    pipelineSvc.refresh(this.pipeline);
+    await pipelineSvc.refresh(this.pipeline);
```
billy-audere: Good catch.

PR #563, opened 2020-01-22
Setting up install steps for FluApi in CircleCI

billy-audere:
Adds jobs for building FluApi images and deploying them to ECS that can be invoked via the `build-service` script on developer workstations using a CircleCI API token. This should create a more repeatable and auditable deployment. All artifacts should be created in the same environment and we should have logs of deployments once they've been kicked off.

A few notes and more pressing concerns:

* Instead of tagging and deploying containers with a latest tag the script will explicitly use the commit as the tag. I think this will be convenient for determining which commit is deployed in prod and staging & better differentiates what is published.
* Build FluApi and deploy staging should probably be a single target. It is possible they will have different commits if an explicit commit is not provided.
* Since we're using CircleCI 2.0 we can't use orbs (CircleCI plug-ins) which would be helpful if we use CircleCI to automated AWS interactions. Unfortunately version 2.1 of the configuration removes the ability to invoke jobs the way we are doing. The new alternative, pipeline parameters & conditional workflows, is a notable change and relatively complicated in comparison.
* There are no checks in place presently about whether a given commit tag was already pushed to ECR. I believe trying to rebuild the same commit would result in a slightly unintuitive error since the user likely has no idea what commits are in ECR.
* There is an access key that has been generated with access to perform the requisite ECR actions. The details will need to be added to the CircleCI configuration.

Adding in notifications to Slack and pruning ECR images to a reasonable number seem like useful quality of life improvements to add on after we decide if running the deployment in CircleCI is right for us.


.circleci/config.yml
```diff
@@ -306,6 +345,54 @@ jobs:
...
+          command: |
+            TAG=$(git rev-parse HEAD)
+            docker build -t fluapi:$TAG -f FluApi/docker/api/Dockerfile .
+            scripts/manage-service publish fluapi $TAG
+
+  deploy_fluapi_staging:
+    <<: *defaults
+    steps:
+      - run:
+          name: Checkout Specific Commit
```
rightparen: Is this here because manage-service might be broken in master?

scripts/build-service
```diff
@@ -0,0 +1,83 @@
...
+ACTION="$2"
+
+shift
+shift
+
+while [[ $# -gt 0 ]]
+do
+key="$1"
+
+case $key in
```
rightparen: TIL the word following `case` does not undergo word-splitting.  I guess that makes sense.

.circleci/config.yml
```diff
@@ -306,6 +345,54 @@ jobs:
...
+      - run:
+          name: Deploy Staging Server
+          command: |
+            TAG=$(git rev-parse HEAD)
+            scripts/manage-service stage fluapi $TAG
+
+  deploy_fluapi_prod:
+    <<: *defaults
+    steps:
+      - run: *install_aws_cli
```
tpaik18: It looks like you can deploy staging from a specific commit but not prod? Wouldn't it be better to be consistent?

billy-audere: We can amend but I was trying to enforce a rigor that any production deployment must be done in staging first.  Prod deployments only promote staging, they can not be a custom build.

billy-audere:
@tpaik18 I've modified the script so that building the image and deploying can be done in a single step.

.circleci/config.yml
billy-audere: I have dropped this step.  I'm not sure why I felt it might be useful to do for staging but not prod.  For now I'll just assume that we always want to use the most recent manage-service.

PR #564, opened 2020-01-22
Returning a specific error message when barcodes do not match

billy-audere:
This will help us distinguish between a request that may have encountered a 404 for another reason & one that successfully attempted to match but did not find the corresponding data. The client should be able to surface different messaging when it is known that the user's barcode is causing the error response.


billy-audere:
@audererob Updated as requested in case we ever decide to handle this.

PR #569, opened 2020-01-24
terraform: datasci 2-factor

rightparen:


rightparen:
Enables DataScientists to manage their MFA and requires it for other operations, just like other roles.

PR #570, opened 2020-01-24
Setting env var in combined deploy step for FluApi on CircleCI

billy-audere:
Each step in the job is a command in a different environment. Environment variables do not propagate. The value needs to be reset for each command.

This may be difficult to understand from the logs because the value undergoes secret masking in the build logs. The commit is passed to the build job (via curl) as an environment variable and all such environment variables are masked by default. We likely do not want to change this behavior but it is something to be aware of. As a result we generally can't see the exact commit being built in the build logs.


billy-audere:
`buildInfo` was not updated from the script because the correct tag wasn't being used in the deploy step & as a result the new image was not deployed.

PR #571, opened 2020-01-24
Limiting reserved executions for Chills MTL Lambda to 1

billy-audere:
MTL files appear to be uploaded in bulk. We can get 200+ files in a short timeframe. Each of these new objects created in S3 will notify the Lambda function to kick off a new file import. Although there is a lock on the endpoint in FluApi the incoming requests seem to overwhelm the FluApi container which then restarts once it fails the ELB healthcheck. This crash/restart leads to a brief period of downtime for the API while ECS cycles the service.

With this change the Lambda should never spin up more than 1 concurrent instance and the API should be better able to handle the request load for MTL import. As best I can tell the Lambda will reject incoming notifications. We do not have a DLQ configured so we won't be able to see this directly. There is some likelihood that new files uploaded while the Lambda is running may not be picked up in a timely fashion although they would be picked up on subsequent runs of the job.

With a little more work the notification from S3 could be funneled into a SNS or SQS topic so that we have more control of how those messages are read & can throttle without failing or losing incoming notifications. As well it might be possible to setup Express in cluster mode to better handle the incoming request load. A better diagnosis of why the Express API is unresponsive would benefit us in choosing the most appropriate action.

Hopefully in the interim this will prevent the API from needlessly restarting.


billy-audere:
These outline a bit more detail on how reserved concurrency works with Lambda functions:

https://aws.amazon.com/blogs/compute/managing-aws-lambda-function-concurrency/
https://docs.aws.amazon.com/lambda/latest/dg/scaling.html

billy-audere:
Also of note I initially assumed (incorrectly) that the restart from the ELB healthcheck was just a new deployment.  The relevant logs to look for in the ECS service events:

```
dfb3a50f-4286-47a6-be03-d0be684db687
2020-01-24 11:18:06 -0800
service fluapi-prod (port 444) is unhealthy in target-group flu-prod-api-internal due to (reason Request timed out).
750686c3-68a8-45df-860c-c846dd73f983
2020-01-24 11:18:06 -0800
service fluapi-prod (port 443) is unhealthy in target-group flu-prod-api-public due to (reason Request timed out).
```

PR #574, opened 2020-01-24
FEV-1575, FEV-1576: post-deploy tests

smeds1:
This PR adds post deploy tests to make sure that Chills barcode verification is working and Cough giftcards are available. I tested the changes in staging, and everything worked. I was unable to test in prod because I do not have the secret keys. @tpaik18 When this PR looks good to merge, I will need help setting up the environment variables in Circle.


smeds1:
@tpaik18 I moved everything into one file and figured out the correct access keys for both staging and prod. Is it ok to put those keys directly into Circle env variables?

tpaik18:
@smeds1 I think it's okay to set the keys as Circle env variables. 
Before we merge this though, I think we should consider whether we are exposing too much (e.g., ability to access a gift card? ability to get through Chills in demo mode?) by checking in these tests into source. @balderdash have any opinions?

balderdash:
I would suggest moving the docIds and barcodes (or at least the prod ones) to circle env variables as well, but otherwise the test doesn't expose anything that can't already be gleaned from our public source code.

scripts/test-fluapi-post-deploy
```diff
@@ -19,21 +19,71 @@ function fail() {
...
+  expected_result="{\"email\":\"t**t@auderenow.org\",\"emailHash\":*"
+  print_test_outcome "${actual_result}" "${expected_result}"
+  return 0
+}
+
+function do_cough_giftcard_test() {
+  local base_url="$1"
+  echo -e "${YELLOW}Testing /api/cough/giftcardAvailable${NC}"
+  if [[ $base_url == $staging_BASEURL ]]; then
+    data="giftcardRequest=\\{\"secret\":\"SECRET_KEY\",\"denomination\":25,\"isDemo\":false,\"docId\":\"$COUGH_TEST_STAGING_DOCID\",\"barcode\":\"$COUGH_TEST_BARCODE\"\\}"
```
billy-audere: Are we able to move the secret key out of the url parameters?

smeds1: I believe that would involve changing @balderdash's giftcardsAvailable endpoint from a GET to a POST. Is that something we want to do?

tpaik18: Can we change GET to POST without breaking the Cough client?

smeds1: @tpaik18 Probably someone can, but I'm not sure that person should be me.

billy-audere: I don't think this change is in scope for this PR.  Please ignore!

tpaik18:
@smeds1 Before merging, can you please re-write the git commits in this branch so the old commits are wiped away?

PR #577, opened 2020-01-28
Adding CloudWatch alarms for load balancer unhealthy hosts

billy-audere:
This should be able replace the curl that is running on Sunflower.


tpaik18:
Does public = prod and internal = staging? If so, then could we use "prod" and "staging" since those terms are more commonly used?

billy-audere:
@tpaik18 FluApi has two sets of endpoints.  One set is only accessible from within AWS.  Those endpoints are behind a separate load balancer that is not public.  We would have 4 alarms if we applied this change:

staging + public
staging + internal
prod + public
prod + internal

tpaik18:
@billy-audere Oh, thanks for educating me. Will it be easy to see from the alarm whether it's prod or staging with the problem? If so, my concern is addressed.

rightparen:
@tpaik18 the alarm name mentions staging vs. prod, so this distinction should show up in the alert.

PR #578, opened 2020-01-28
FEV-1574: Add Slack notifications for deploys

smeds1:
This PR attempts to add Slack notifications for deploys. I wasn't sure how to test the code locally, so there are probably some issues that need to be fixed up before merging.


scripts/manage-service
```diff
@@ -212,6 +214,7 @@ case "$subcommand" in
     ecr_login
     ecs_deploy "$service_id" "$service" "$repos" "$default_tag" "$new_tag" "$@"
 
+    curl -X POST --data-urlencode "payload={\"text\": \"Deploy succeeded!\", \"icon_emoji\": \":large_blue_circle:\"}" $INFRA_SLACK_WEBHOOK
```
billy-audere: Let's add the service argument, environment, and commit to the message if possible.

scripts/manage-service
```diff
@@ -184,6 +185,7 @@ case "$subcommand" in
     repo="$1"
     tag="${2:-latest}"
 
+    curl -X POST --data-urlencode "payload={\"text\": \"$(whoami) is deploying $repo $tag\", \"icon_emoji\": \":large_blue_circle:\"}" $INFRA_SLACK_WEBHOOK
```
billy-audere: When this runs on CircleCI let's try to get the identity of the caller similar to how the app build is configured:

https://github.com/AudereNow/audere/blob/master/FluStudy_us/android/fastlane/Fastfile#L84

I assume we'd want to check to the CircleCI env variable and fallback to the local user if that environment variable isn't present.

tpaik18: Right -- CIRCLE_USERNAME should give you what you need.

scripts/manage-service
smeds1: @billy-audere What do you mean by environment? I thought you were talking about staging/prod, but that seems to be included in $service.

PR #579, opened 2020-01-28
Post deploy version check

smeds1:
This PR checks to see that the hash returned in /about is the latest commit. Do we need to worry about deploying commits other than the latest one?


scripts/test-fluapi-post-deploy
```diff
@@ -38,11 +38,11 @@ function do_api_test() {
   return 0
 }
 
-function do_about_test() {
+function do_version_test() {
   local base_url="$1"
   echo -e "${YELLOW}Testing /about${NC}"
   actual_result="$(curl -sS "${base_url}/about")"
-  expected_result="{\"name\":\"nodejs-web-service\"*"
+  expected_result="*\"hash\":\"$(git rev-parse HEAD)\"}"
```
billy-audere: This logic won't always yield the correct revision.

When we do a prod deployment we only deploy what is currently marked as staging.  We don't checkout to a specific commit and may not be up to date with master.  If we want the revision we'd have to check the tags from ECR for the image that is already marked as staging.

smeds1: @billy-audere Is the ECR tag something we're already using in this script, or do I have to go and get it from somewhere? I see the tag from my deploy yesterday is now "prod, staging, commit_hash". Will it always be in that format for parsing purposes?

billy-audere: We always expect that to be the format we use for FluApi.

We are not using this anywhere else in the script.  The best way I can think of to retrieve it would be something like:

* `aws ecr describe-images --repository-name fluapi --image-ids imageTag=staging` to get the image description.
* Then parse the image description with `jq` to get the specific commit tag.

We would also need to pass in the environment being deployed to, which is also used as a tag on the image.  The environment tag is used to anchor deployments.

It may make sense to move testing to a different subcommand within `manage-service` which we can call from the CirlceCI config & can therefore more directly pass parameters or set environment variables for testing.

scripts/test-fluapi-post-deploy
```diff
@@ -38,11 +38,13 @@ function do_api_test() {
...
 }
 
-function do_about_test() {
+function do_version_test() {
   local base_url="$1"
   echo -e "${YELLOW}Testing /about${NC}"
   actual_result="$(curl -sS "${base_url}/about")"
-  expected_result="{\"name\":\"nodejs-web-service\"*"
+  aws_ecr=$(aws ecr describe-images --repository-name fluapi --image-ids imageTag=staging)
+  commit="$(echo $aws_ecr | jq -r ".imageDetails[]|.imageTags[2]")"
```
smeds1: @billy-audere Are you saying that this will only work if deploying to prod and the tag will be different if deploying to staging?

billy-audere: The issue is that it will always be synced to head.  So for instance if we put out a build to staging, do some directed testing on it & in the interim more commits are added to master it'll report a revision that is not tagged in ECR and is inaccurate when deploying prod.

scripts/test-fluapi-post-deploy
```diff
@@ -43,7 +43,15 @@ function do_version_test() {
...
   aws_ecr=$(aws ecr describe-images --repository-name fluapi --image-ids imageTag=staging)
-  commit="$(echo $aws_ecr | jq -r ".imageDetails[]|.imageTags[2]")"
+  image_details="$(echo $aws_ecr | jq -r ".imageDetails[]|.imageTags")"
+  for tag in $image_details; do
+    if [[ $tag =~ ^\"([A-Za-z0-9]{40})\" ]]; then
+        commit=${BASH_REMATCH[1]}
+    fi
+  done
+  if [[ -z "$commit" ]]; then
+    fail "Could not find commit in erc image details"
```
billy-audere: erc -> ECR

PR #582, opened 2020-01-30
FEV-1573: Migrating Slack notifier to a new project

billy-audere:
The previous GitHub project we were using did not work well with the Node 10.x runtime in Lambda. Since we have a hard requirement to update our Lambda functions to Node 10.x or 12.x I've updated this to a new library. The newer library has a simpler deployment so I've modified init-local and removed the duplicate init-local from the cron Lambda module.


billy-audere:
This should address issues described in FEV-1573

billy-audere:
I've also updated the parameter store value for the Slack webhook as part of this (needed to prepend the protocol for this package to succeed in making a notify request).

PR #583, opened 2020-01-31
Update RDT reader to remove/relax IPRD filters, and send preview fram…

jennya:
…e data to the server


FluStudy_us/android/app/build.gradle
```diff
@@ -39,7 +39,7 @@ android {
     manifestPlaceholders = [
       'appAuthRedirectScheme': 'host.exp.exponent'
     ]
-    minSdkVersion 22
+    minSdkVersion 21
```
tpaik18: Was there a reason this was necessary?

jennya: Oh no this was just so that I could run the app on the Samsung Galaxy Note. I will revert

FluStudy_us/src/ui/components/flu/AndroidRDTReader.tsx
```diff
@@ -491,6 +505,30 @@ class AndroidRDTReader extends React.Component<Props & WithNamespaces, State> {
     this._interpreting = true;
   };
 
+  _logPreviewFrameData = async (args: RDTCapturedArgs) => {
+    const rdtResult = rdtCapturedArgsToResult(args);
+
+    const sampleRate = getRemoteConfig("previewFramesSampleRate");
+    rdtResult.photoUploaded =
+      sampleRate > 0 && args.previewFrameIndex % sampleRate == 0;
```
audererob: If I'm understanding the code correctly, I think you could return true/false from uploadPreview (depending on whether it was queued vs deleted) and avoid having this logic in two places.

PR #586, opened 2020-02-04
Updating CircleCI deployment notification for FluApi

billy-audere:
* Removing commit information as we don't have the current FluApi commit readily available during deployment
* Adding build URL to all notifications
* Moving first deploy notification to the deploy command


billy-audere:
@tpaik18 I had to make some changes to the post-deployment ready check.  The new check should verify that a service has been switched over and is healthy on the load balancer.  This is more inline with what we hope for as the end status for the deployment, otherwise we're prone to querying/testing the old service before it is taken down.

Note that this information is not readily available by querying ECS, especially at the task level.  Rather we must parse the service logs to determine whether this has occurred.  We filter by timestamp and wait for the given timeout for such an event to occur.

Also note that this took about 3 minutes to finish when I tested in staging.

PR #587, opened 2020-02-04
Moving MTL import to a cron schedule

billy-audere:
Currently the S3 notification triggers frequently in short intervals as Evidation uploads status updates in batch at two mostly fixed times. This invokes the job dozens of times. There are some database resource timeouts that also occur intermittenly while running those invocations sequentially. This moves the job to trigger twice per day instead of subscribing the S3 notifications. This should reduce the number of invocations that happen daily, reducing wasted work. As well we can monitor the database timeouts after this change to see if there may be an underlying issue separate from db contention.


billy-audere:
Since we are filtering [here](https://github.com/AudereNow/audere/blob/6cf1e3f689f4fdfdce1eb5796728d8130bc40d96/FluApi/src/external/evidationMTLClient.ts#L80) there isn't an immediate need to implement a cursor to resume reading files.  It should be safe to invoke the endpoint as-is & yield the desired results.

Adding a cursor to read from would be more efficient and prevent issues with the dynamic SQL that results from checking for processed files if the file list grows too large (we are susceptible to issues if the # of files uploaded in the last 3 days gets into the thousands).

PR #590, opened 2020-02-05
FEV-1572: Prune images from repos in ECR

smeds1:
The goal of this PR is to set a rule for pruning ECR images so that they do not build up indefinitely and cost us unnecessary money. Here are some questions I have:

1) Is this code in the right file?
2) Does the "aws_erc_repository" resource look like correct terraform syntax?
3) Is there a way to specify the same rule for all repos, or do I need to write out similar code for the other three ECR repos?


billy-audere:
The syntax is correct.  To check the syntax you can run `terraform plan`.  This is the output (the proposed changes) from this branch:

```
An execution plan has been generated and is shown below.
Resource actions are indicated with the following symbols:
  + create

Terraform will perform the following actions:

  + aws_ecr_lifecycle_policy.fluapi_ecr_repo_policy
      id:             <computed>
      policy:         "{\n    \"rules\": [\n        {\n            \"rulePriority\": 1,\n            \"description\": \"Expire older images when image count is more than 100\",\n            \"selection\": {\n                \"tagStatus\": \"any\",\n                \"countType\": \"imageCountMoreThan\",\n                \"countNumber\": 100\n            },\n            \"action\": {\n                \"type\": \"expire\"\n            }\n        }\n    ]\n}\n"
      registry_id:    <computed>
      repository:     "fluapi"

  + aws_ecr_repository.fluapi_ecr_repo
      id:             <computed>
      arn:            <computed>
      name:           "fluapi"
      registry_id:    <computed>
      repository_url: <computed>


Plan: 2 to add, 0 to change, 0 to destroy.
```

Adding the fluapi repo may fail since it already exists but we can run Terraform commands to assign the existing repo to the Terraform resource if that occurs.

I don't see a way to specify a global lifecycle policy.  Since we have created the repos dynamically and they aren't managed by Terraform we can't iterate over them and I don't think AWS supports attaching a policy to multiple repos (or a wildcard of repos).

PR #596, opened 2020-02-10
FEV-1660 Show the "multiple pics" alert only if previewFramesSampleRate > 0

tpaik18:
Also, follow-on fix for FEV-1659. It wasn't working properly because I forgot to `await`.


FluStudy_us/src/ui/components/flu/AndroidRDTReader.tsx
```diff
@@ -145,9 +145,11 @@ interface State {
 }
 
 class AndroidRDTReader extends React.Component<Props & WithNamespaces, State> {
+  _sampleRate: number = getRemoteConfig("previewFramesSampleRate");
```
audererob: Can be const

FluStudy_us/src/ui/components/flu/AndroidRDTReader.tsx
```diff
@@ -145,9 +145,11 @@ interface State {
...
+
   state: State = {
-    spinner: false,
-    showCamera: false,
+    spinner: this._sampleRate > 0 ? false : true,
```
audererob: Nit: could be spinner: this._sampleRate === 0,
(arguable if that's more/less readable)

FluStudy_us/src/ui/components/flu/AndroidRDTReader.tsx
tpaik18: For some reason when I add `const` to the front Prettier strips it away.

rightparen: @tpaik18 fields use `readonly`, locals use `const`.

PR #598, opened 2020-02-10
Process frames after user agrees to start scan

jennya:
Changes the control flow for the reader to first open the RDT reader and initialize the camera, then (if sample rate > 0) show the 'start scan' dialog, then finally begin processing the frames. This way the user can see their strip underneath the dialog before they hit start (but we are not capturing anything before they explicitly begin the scan)


jennya:
Ah you're right -- it will start as soon as they hit this screen. I will move setTimer to _startRecordingFrames()

PR #602, opened 2020-02-12
Upload preview frame data as separate files

jennya:
Removes the huge chunks of preview frame data from redux.

Now for each RDT reader session, writes and uploads a new file called `preview_info_<uid>.<install id>.json` to the firebase `photos/` storage. Within each survey doc, we now will have an array of each preview_info filename.


FluStudy_us/src/ui/components/flu/AndroidRDTReader.tsx
```diff
@@ -274,13 +275,19 @@ class AndroidRDTReader extends React.Component<Props & WithNamespaces, State> {
...
+  _saveAndClearPreviewFrames = async () => {
     if (this._previewFrames.length > 0) {
-      dispatch(appendPreviewSeries(this._previewFrames));
+      const filename = "preview_info_" + (await newUID()) + ".json";
+      const path = FileSystem.documentDirectory + filename;
+      await FileSystem.writeAsStringAsync(
+        path,
+        JSON.stringify(this._previewFrames)
+      );
+      uploadFile(filename, path, true);
```
tpaik18: @jennya Do these files get deleted off the file system after upload?

jennya: Yeah, that's what the ambiguous `true` flag is saying -- delete this file after upload

PR #603, opened 2020-02-14
FEV-1693: update evidation trigger date

smeds1:
Evidation found an error in the data they have been giving us. The created a new tab in the gsheet we have already been using (see picture below). I would like to import the "external id" and "Flu Kit Trigger Date" columns from this sheet instead of the old one.  

image


FluApi/src/external/evidationTriggersClient.ts
```diff
@@ -32,15 +32,15 @@ export class EvidationTriggersClient {
...
     });
 
-    const sheet = spreadsheet.data.sheets[0];
+    const sheet = spreadsheet.data.sheets[1];
     const data = sheet.data[0];
 
     const kits = data.rowData
-      .slice(1)
-      .filter(row => row.values[1].formattedValue != null)
+      .slice(2)
```
balderdash: Nit: Consider using constants for these row and column numbers, e.g. `HEADER_ROW_COUNT = 2`, `FLU_KIT_TRIGGER_DATE_COL = 4`, etc.

smeds1: Good call. If they actually intend to maintain this new sheet, I will set up some constants and import the other columns.

PR #604, opened 2020-02-20
FEV-1692: add rdt preview series info tables

smeds1:
The purpose of this PR is to extract RDT preview series information from Firebase storage and insert it into database tables for easier analysis. The table rdt_preview_frames contains one entry per preview frame logged during a capture attempt. The column series_index logs which rdt capture attempt the frame belongs to (0 for first attempt, 1 for second attempt, etc). 

This code works locally except that it doesn't upsert. If the table already has an identical row, it inserts another copy. @billy-audere Can you help me figure out why the upsert is not working the way that it does for the surveys table?

Note: I haven't put in the cron job yet. I will add that in a separate PR once I'm confident that this code works.


billy-audere:
@smeds1 Within Sequelize upsert is implemented as exception handling to a unique constraint violation.  See [here](https://github.com/sequelize/sequelize/issues/4132) (you could inspect the SQL generated locally for an upsert by debugging).  In order for this to work you need a unique constraint on the table.

For `rdt_preview_frames` you could consider:

```
docid: unique(column(Sequelize.STRING)),
```

smeds1:
@billy-audere Unfortunately docid is not unique since there are many preview frames for each docid. I could create some kind of unique identifier that is docid-series-frame. Would that be too weird?

billy-audere:
@smeds1 You should be able to create a composite unique constraint on whatever combination of columns would be unique, such as docid + series_index + frame_index, etc.

smeds1:
@billy-audere Thanks. Adding the unique constraint fixed my upsert problems.

PR #605, opened 2020-02-24
Fetch data during import instead of within the endpoint constructor.

billy-audere:
Currently the newest data for CDC import is not being imported in production. This may be because the data from the previous week still exists in memory, and is cached, since the service has not been redeployed recently.


smeds1:
Looks good now.

PR #606, opened 2020-02-25
Explicitly setting the database CA to 2019

billy-audere:
This change is required by AWS. It should be applied in the next maintenance window, or can be applied immediately using apply_immediately = true. I have chosen to wait for the next maintenance window to avoid any unnecessary downtime.


billy-audere:
I do not believe we have to update the trust store with the new CA to maintain connection integrity but will test it out in staging first.

PR #608, opened 2020-02-29
Importing Chills RDT preview data one file at a time

billy-audere:
* Splits the fetch of preview data into two functions: one listing unprocessed doc ids & one to fetch & interpret the preview frames within a given image doc.  I focused on minimizing the amount of information held in memory by the API server.  This proposes to use the SQL server for filtering the survey documents so we are only handling a string/path for the image doc id instead of fetching the entire corpus of survey documents and doing filtering in memory.  We were previously fetching all the images and similarly holding them in memory as we iterated over them.  Instead we now process a single image in memory at any given time.
* I queried production and there were 1178 doc ids referencing an image from the series.  Within that all 1178 ids were unique.  Since we're using a two part id I believe that the doc id should always be unique and we don't generate duplicate frames.  As a result I've removed the unique constraint on doc id + series index + frame index from `rdt_preview_frames` and replaced it with a simpler unique index on doc id.  This simplifies tracking what has been processed but can be updated if the assumption of uniqueness for the doc id is incorrect.
* Corrects a number of variable names to use camel casing


smeds1:
@billy-audere This branch runs into some errors when I try to run it locally (operator does not exist: record ->> unknown). Also, I'm skeptical that we can make docid a unique constraint. A single failed capture generates ~80 preview frames, and each frame is a row in the database. Wouldn't that result in 80 rows with the same docid?

PR #609, opened 2020-03-03
Attempt to save preview frames only from existing photos

billy-audere:
If there is a survey document uploaded but the corresponding photo has not yet arrived, or will not arrive due to transport errors such as the user going offline etc, then we do not attempt to fetch that photo. It will be picked up when both the survey and photo exist.


billy-audere:
@smeds1 You should be able to manually delete the corresponding photo row to test.

PR #618, opened 2020-04-17
FEV-1807: add photo finished timestamp to fever metrics report

smeds1:
Monica and Ivan at SFS would like an additional column in the report that shows when the user uploaded the image from the first test. I used the "cleanFirstTest" screen since it immediately follows image upload (which some users skipped).


smeds1:
Oh boy

PR #631, opened 2020-05-14
Create new test on home if enableProfile is false

ethantmanns:
If `enableProfile` is false, navigating to the root of the app will create a new test run under the `UNSET_PROFILE` profile ID and navigate to that test run.

One thing to consider is clicking on the logo will create a new test. Should we consider where the logo in the navbar links to?


OpenRDTReaderWeb/src/utils/AppConfig.ts
```diff
@@ -17,6 +17,8 @@ const MasterConfig: AppConfig = {
   photoUploadEnabled: true,
 
   staticPhotoMLRecognitionEnabled: true,
+
+  enableProfiles: true,
```
ethantmanns, line 21: Is this the default we want? I assume so. I just had the override set locally to false to test

OpenRDTReaderWeb/src/components/App/DashboardApp.tsx
```diff
@@ -4,24 +4,65 @@
...
+        startingPath = newTestPath;
+      }
+    }
+
+    setStartingPath(startingPath);
+  };
+
+  useEffect(() => {
+    getStartingPage();
+  });
```
ethantmanns, line 64: I'll add `[auth, config]` here

yrichard:
Ah, yes, I think the logo should create a new test run as well.
Could we go to a page which does only this: 
- Create a new testrunid
- redirect to the start page for testruns.

OpenRDTReaderWeb/src/utils/AppConfig.ts
```diff
@@ -17,6 +17,8 @@ const MasterConfig: AppConfig = {
   photoUploadEnabled: true,
 
   staticPhotoMLRecognitionEnabled: true,
+
+  landingPage: "HOME",
```
yrichard, line 21: let's put NEWTESTRUN here.

PR #632, opened 2020-05-15
[OpenRDT Web] Remove NavAuthCache

mathewsb:
Since AudereNow/openrdt-internal#222, we no longer render `NavBar` when
`loading` is true.  So this `NavAuthCache` code does nothing and can be
removed.


yrichard:
Although, this is true for this configuration we did for NHS, I am not sure we will stay with a navbar-less logged out experience on the long run. Is it necessary to remove this when we will introduce redux?

mathewsb:
No, this is isn't needed for the Redux diff.  It's just that I was trying to test this code, and I found that it was unreachable...

Okay, I'll just abandon for now

PR #640, opened 2020-05-15
USC-830 Make server assets possible

auderephilip:
This introduces a ServerAsset component which allows images to be loaded from the server when needed. Puts a div placeholder in place until the image is fully loaded.

Here's it working on the setup screen, as a test, using the following code in TestRunConstants:
```
      <ServerAsset
        height={180}
        width={290}
        alt="Image of a kit box and test vial in a bag."
        src="rdtCassette.png"
      />
```
image

And here are all the assets that we were provided in the zip file today, in Firebase Storage:
image

Huge thanks to @yrichard for teaching me functional React components that require async setups.


yrichard:
nice!

PR #642, opened 2020-05-15
USC-814 Add sign-in content

auderephilip:
Just added a sentence here.

Tested by verifying that logging in still works:
image


OpenRDTReaderWeb/src/components/Login/Login.tsx
```diff
@@ -50,6 +50,12 @@ export default () => {
           submitLabel="Login"
           error={error}
         >
+          <div>
+            The COVID-19 Antibody Home Testing App is offered by the NHS, and
+            enables you to order and conduct a COVID-19 antibody test from the
+            comfort of your home.
+          </div>
+          <br />
```
auderephilip, line 58: LMK if <br> is considered lame. I figure we're just temporarily putting something here, and that there will be a richer design later.

yrichard: Yeah, I think we can allow little shortcuts like that, no worries.

PR #649, opened 2020-05-17
Add breaks in positive result message

ethantmanns:
Before:
image

After: 
image


yrichard:
@ethantmanns, this is acceptable but for reference, if you want something to go on a new line and take the full width, you usually need to wrap it into something with display: block, such as a 'div' for example.

PR #656, opened 2020-05-19
USC-765 close datepicker

ethantmanns:
Auto-close datepicker when a date is selected
datepicker


OpenRDTReaderWeb/src/components/ui/Questionnaire/QuestionnaireInputDate.tsx
```diff
@@ -67,10 +68,15 @@ const QuestionnaireInputDate = (props: QuestionnaireInputProps) => {
         },
       };
       onChange(event);
+      setPickerOpen(false);
     },
     [onChange, question.name]
   );
 
+  const togglePicker = () => {
```
yrichard, line 76: use "useCallback" here to cache the function between renders.

PR #662, opened 2020-05-20
Convert questionnaire select to material-ui

ethantmanns:
Use Material-ui select instead of native select for more consistent styling


OpenRDTReaderWeb/src/components/ui/Questionnaire/QuestionnaireInputSelect.tsx
```diff
@@ -31,43 +47,49 @@ const QuestionnaireInputSelect = (props: QuestionnaireInputProps) => {
...
+      const dummyEvent: DummyEvent = {
+        currentTarget: {
+          name: question.name,
+          value,
+        },
+      };
+      onChange(dummyEvent);
+    },
+    [onChange, question.name]
+  );
```
ethantmanns, line 66: We had to do something similar for date. I think I'm going to see how the other components handle this as well and if it's easy to share some code there I'll put something up in the later PRs

OpenRDTReaderWeb/src/components/ui/Questionnaire/QuestionnaireInputSelect.tsx
```diff
@@ -31,43 +47,49 @@ const QuestionnaireInputSelect = (props: QuestionnaireInputProps) => {
...
-            ))}
-          </select>
-        </div>
-        {question.icon && (
-          <span className="icon is-small is-left">
-            <FontAwesomeIcon icon={question.icon as IconName} />
-          </span>
-        )}
-      </div>
+      <FormControl error={!!error}>
```
yrichard, line 74: We don't need this, QuestionnaireStandardFormControl already provides this wrapper. You might need to change QuestionnaireStandardFormControlProps.children to be a ReactNode instead of JSX.Element, so that it can accept more than one element as children.

OpenRDTReaderWeb/src/components/ui/Questionnaire/QuestionnaireInputSelect.tsx
```diff
@@ -31,43 +47,49 @@ const QuestionnaireInputSelect = (props: QuestionnaireInputProps) => {
...
+          <MenuItem value="" disabled>
+            {question.placeholder}
+          </MenuItem>
+          {question.selectOptions.map(selectOption => (
+            <MenuItem key={selectOption.value} value={selectOption.value}>
+              {selectOption.name}
+            </MenuItem>
+          ))}
+        </Select>
+        <FormHelperText>{question.placeholder}</FormHelperText>
```
yrichard, line 91: Let's not use this, it's redundant with the placeholder in the selector.

OpenRDTReaderWeb/src/config/example/CustomSelectExample.tsx
```diff
@@ -0,0 +1,74 @@
+// Copyright (c) 2020 by Audere
```
yrichard, line 1: awesome! Thanks :)

ethantmanns:
Screen Shot 2020-05-20 at 2 42 34 PM

OpenRDTReaderWeb/src/components/ui/Questionnaire/QuestionnaireInputSelect.tsx
```diff
@@ -31,43 +44,48 @@ const QuestionnaireInputSelect = (props: QuestionnaireInputProps) => {
...
-      <div
-        className={cx({
-          control: true,
-          "has-icons-left": !!question.icon,
-        })}
+      <Select
+        value={question.value || ""}
+        displayEmpty
+        className={classes.select}
+        inputProps={{ "aria-label": "Without label" }}
```
yrichard, line 75: Should we pass question.label here?

ethantmanns: Actually, do we need this at all? I might just remove

ethantmanns: Ah now I get what it's for

PR #663, opened 2020-05-20
Replace checkbox with material-ui

ethantmanns:
Screen Shot 2020-05-20 at 2 20 24 PM
Screen Shot 2020-05-20 at 2 20 17 PM


OpenRDTReaderWeb/src/components/ui/Questionnaire/QuestionnaireInputCheckbox.tsx
```diff
@@ -4,60 +4,59 @@
...
+    ) => {
+      const value = event.target.checked ? "true" : "false";
+
+      const dummyEvent: DummyEvent = {
+        currentTarget: {
+          name: question.name,
+          value,
+        },
+      };
+      onChange(dummyEvent);
```
yrichard, line 39: do we need this? If you look at the original code, I just added the value to the existing target.
`
        function(ev) {	   
            ev.target.value = ev.target.checked ? "true" : "false";	        
            onChange(ev);	        
  }
`

ethantmanns: Typescript still gets upset since it's different between `ChangeEvent<{ name?: string; checked: boolean }>` and `ChangeEvent<element>`

yrichard: how weird, ok then...

PR #664, opened 2020-05-20
OpenRDT Web: Composable List Component

yrichard:
In order to have a more configurable, I will create a set of "Composable" components that can be composed like standard react components, but also can take a more declarative approach.

For this example, the ComposableList accepts children (like a list of "li"), and an array of strings (or fbts) that will be converted into "li" elements.

I will do this for paragraphs, title, subtitle, and timer as well.

While I was at it, I added fbt wrappers for all the strings I touched.


OpenRDTReaderWeb/src/components/TestRun/TestRunConstants.tsx
```diff
@@ -134,13 +144,27 @@ const testrunSteps: IncompleteStepsDefinition = {
...
         </div>
-        <ul className={getStepStyle().bulletList}>
-          <li>white package (test cassette)</li>
-          <li>pipette (thin tube)</li>
-          <li>lancet (finger prick tool)</li>
-          <li>small bottle of buffer</li>
-          <li>bag for kit disposal</li>
-        </ul>
+        <ComposableList
+          items={[
```
yrichard, line 148: This is just to test how arrays would work.

OpenRDTReaderWeb/fbt/.source_strings.json
```diff
@@ -19,9 +19,9 @@
     "AnfZf5e3aLc7f/AkqiQXaw==": "Sign In"
```
yrichard, line 19: generated

OpenRDTReaderWeb/fbt/.src_manifest.json
```diff
@@ -1 +1 @@
-{"fbt/.enum_manifest.json":["src/components/App/DashboardApp.tsx","src/components/Login/AuthenticationCard.tsx","src/components/Login/Login.tsx","src/decs.d.ts"]}
\ No newline at end of file
+{"fbt/.enum_manifest.json":["src/components/App/DashboardApp.tsx","src/components/ComposableComponents/ComposableList.tsx","src/components/Login/AuthenticationCard.tsx","src/components/Login/Login.tsx","src/components/TestRun/TestRunConstants.tsx","src/decs.d.ts"]}
```
yrichard, line 1: generated

PR #669, opened 2020-05-22
[OpenRDT Web] Increase the resolution of photos from the inline camera

mathewsb:
The way that `jslib-html5-camera-photo` tries to implement "max resolution"
doesn't actually work on iOS Safari.  (The uploaded photo ends up being
640x480 pixels.)

Change to using the `idealResolution` prop instead.  This seems to work on
all browsers I've tried:
* MacOS Chrome and Safari: I get 1280×720 pixels.  (This is the max supported by the webcam.)
* iOS Safari: I get 1673×2240 pixels.  (This is not the maximum for the camera.)

(I had to work around some issues with `react-html5-camera-photo` to make
this work.  I don't think this is a great package, we might want to roll
our own solution.)


yrichard:
okidoki. Looking at the stats, this package is the best (aka only one) of its kind on npm. It might be better to just fix their bugs rather than roll our own? Or at least limit ourselves to the rewriting the react wrapper around jslib-html5-camera-photo. The browser APIs to support cameras and video streaming are super fragmented, so I would worry about reaching quickly the same level of quality on our own.

PR #670, opened 2020-05-22
OpenRDT Web: Make test run steps be more declarative.

yrichard:
As we want to allow for having multiple configurations, and potentially having those configurations on the server, I am making some changes to our client side definitions of "TestRuns".
For this, I am introducing a "contentDescription" field which will be an array of "Declarative Components". 

A Declarative Component is a chunk of UI that can be a title, paragraph, list, asset, timer or photo uploader. Note that those can be used outside of the "TestRuns", so going forward we should make sure we use them to render all of our titles, paragraphs, lists etc...

Props for those components can be defined in the config and will be passed at render time. 
For translation, I am currently wrapping the strings in FBTs, but obviously, if we generate the config server side, then the server will need to pass translated strings.

An important TODO here, is that the server will most probably want to pass some complex strings with links, bolded text, etc... So we will need to add support for that eventually.


OpenRDTReaderWeb/src/components/TestRun/TestRunSteps.tsx
```diff
@@ -116,11 +117,21 @@ export default () => {
...
+              testRunUID={testRunUID!}
+            />
+          )}
+          {currentStepDescription.ContentComponent && (
+            <currentStepDescription.ContentComponent
+              setStepReady={setStepReady}
+              submitUrl={nextPath}
+              testRunUID={testRunUID!}
+            />
+          )}
```
yrichard, line 134: I will clean that up once we're sure we don't want this anymore.

yrichard:
> It looks good to me!
> 
> We do sort of lose some of the benefits of Typescript linting, though, right? We won't know of the type mismatches until we're rending the components with external props?

Yes, temporarily, we'll lose some linting. I need to make WaitStep and others not dependent on props passed from whatever will render DeclarativeContentRenderer, and then I will be able to reintroduce linting and maybe some type of data validation that doesn't rely on typescript.

PR #672, opened 2020-05-26
OpenRDT Web: Make TestRun components independent from their container.

yrichard:
As we go for a more declarative configuration of the TestRun steps, I am adding a "NextStepButton" component that relies on an event subscription model, rather than passing around props.
This allows us to be more flexible with the UI, by making the button part of the step description, we can now add content below the button itself.
This also allows me to get rid of StepDetailComponentProp.


OpenRDTReaderWeb/src/components/TestRun/NextStepButton.tsx
```diff
@@ -0,0 +1,56 @@
...
+    }
+  }
+
+  useEffect(() => {
+    const emitter = getStepReadyEmitter();
+    emitter.on(step, setEnabled);
+    return () => {
+      emitter.off(step, setEnabled);
+    };
+  }, [step]);
```
ethantmanns, line 53: I'm not sure I fully understand how this is working

yrichard: Ah, I would suggest reading this: https://reactjs.org/docs/hooks-effect.html#example-using-hooks-1
Basically useEffect runs after every render, and returns a function that will be run before every render and when the component unmounts. 
So, in this case here, we subscribe to an event "step-ready" when the component will be mounted, and we will unsubscribe when the component is dismounted.
If we weren't doing this, then some "step ready" event might trigger "setEnabled" for components that don't exist in the page anymore.

PR #674, opened 2020-05-27
Update text and display for upload path

ethantmanns:
Changed the text on the upload path to also say "Start camera". Removed the icon for the upload button. Added a prop for the image upload component to either display a button or a clickable text.

Screen Shot 2020-05-27 at 10 19 50 AM

Screen Shot 2020-05-27 at 10 27 34 AM


ethantmanns:
Closing this since we actually only want to change this text if the device supports capture mode instead of picking a photo to upload

PR #676, opened 2020-05-27
USC-875 USC-876 dynamic upload messages

ethantmanns:
Change the upload messages to "start camera" if the device supports the `capture` attribute on the `input` element, which allows it to use the camera device rather than a file picker.

I did look at using modernizr but there were some hoops we had to jump through to get it working with webpack and we only needed this one function.

## Desktop (no capture support):
Screen Shot 2020-05-27 at 11 34 04 AM
Screen Shot 2020-05-27 at 11 34 15 AM

## Mobile (with capture support):
Screenshot_20200527-115529
Screenshot_20200527-115541


OpenRDTReaderWeb/src/components/FileUploader/PhotoPreview.tsx
```diff
@@ -70,12 +68,12 @@ export default (props: PhotoPreviewProps) => {
...
-          {config.cameraEnabled && (
+          {usedCamera && (
             <fbt desc="Photo preview: camera" project="audere">
               Confirm that your test photo looks good, or choose Re-take photo.
             </fbt>
           )}
-          {!config.cameraEnabled && (
+          {!usedCamera && (
             <fbt desc="Photo preview: upload" project="audere">
               Confirm that your test photo looks good, or choose Re-upload
```
yrichard, line 78: Shouldn't it say "Re-take photo" if inputCaptureSupported?

OpenRDTReaderWeb/src/style/utils.ts
```diff
@@ -61,3 +61,11 @@ export const useAnimateBgAndColor = makeStyles((theme: Theme) =>
 export const toCSSPercent = (x: number): string => {
   return x * 100 + "%";
 };
+
+export const inputCaptureSupported = () => {
+  const testElement = document.createElement("input");
+  testElement.setAttribute("capture", "true");
+  const inputCaptureSupported = !!testElement["capture"];
+  testElement.remove();
```
yrichard, line 69: cache the value so that we only test for it once.

PR #677, opened 2020-05-28
Create customer-specific areas within the OpenRDT website

balderdash:
This PR does a couple things:
* Moves customer-specific pages under `/c/<customer>/...`
* Creates skeletons of customer-specific configs

Things this PR doesn't do yet:
* Actually use the customer configs to render different things for different customer pages


OpenRDTReaderWeb/src/routes/routes.ts
```diff
@@ -257,9 +259,25 @@ export const ROUTE_DEFINITIONS: { [key: string]: DashboardRoute } = {
...
+  if (!customerPrefix) {
+    const match = window.location.pathname.match(/^\/c\/([^/]*)/);
+    if (match) {
+      customerPrefix = match[1];
+    } else {
+      customerPrefix = "nhs"; // TODO(ram): put this in a config somehwere
+    }
+  }
+  const c = `/c/${customerPrefix}${route.path}`;
+  console.log(c);
```
yrichard, line 279: delete

OpenRDTReaderWeb/src/components/Login/PasswordForget.tsx
```diff
@@ -65,7 +65,7 @@ export default () => {
           </div>
         )}
         <div>
-          <Link to={ROUTE_DEFINITIONS.LOGIN.path}>Go back to login</Link>
+          <Link to={getPath(ROUTE_DEFINITIONS.LOGIN)}>Go back to login</Link>
```
yrichard, line 68: Not a needed for this diff but maybe we need a component <AudereLink /> that abstracts the "getPath" call away.

OpenRDTReaderWeb/src/config/AppConfigOverrides.ts
```diff
@@ -12,6 +12,26 @@ import { AppConfig } from "../utils/ConfigTypes";
 const overrides: Partial<AppConfig> = {
   cameraLiveMLRecognitionEnabled: false,
   staticPhotoMLRecognitionEnabled: false,
+  customerConfigs: [
```
yrichard, line 15: this is exciting!

OpenRDTReaderWeb/src/components/App/DashboardApp.tsx
```diff
@@ -57,22 +67,23 @@ const DashboardApp = () => {
                               content={route.seo.description}
                             />
                           </Helmet>
-                          <route.component />
+                          <route.component customerConfig={customerConfig} />
```
yrichard, line 70: This is unused, right now, right? We could use the React Context pattern here instead of passing a prop. That would make the config available easily to any child component. 
https://reactjs.org/docs/context.html
Another solution could be to have a helper function getCurrentConfig which would fetch and store the config for the area based on the url.

PR #679, opened 2020-05-29
OpenRDT Web: Early upload of photos

yrichard:
This diff makes the photo upload to start as soon as you enter the preview screen.
I show a progress bar above the pic, but if the use clicks on the upload button, I switch to the progress circle.
Also I've rearranged the code to only manage the upload aspect on the screen where upload happens.

Tested both with file upload and inline camera.


OpenRDTReaderWeb/src/components/FileUploader/PhotoPreview.tsx
```diff
@@ -3,47 +3,168 @@
...
+      });
+    };
+
+    setProgress(0);
+    setUploadingState(UPLOADING_STATES.BACKGROUND);
+
+    uploadAsync();
+
+    return () => {
+      console.log("cancel");
```
ethantmanns, line 124: Remove?

PR #681, opened 2020-05-29
USC-863 guest test runs

ethantmanns:
As it is, we're able to go through test runs without signing in and they'll be stored under the magic under `UNSET_USER_ID`, but I think there's still some work to be done. Wanted to get some feedback on my approach so far.

Still need to:
- get firebase rules unit tests working (right now assertions are always passing even when they shouldn't)
- See if I can make things less "magic word"y
- Go through and do some cleanup

Screen Shot 2020-05-29 at 1 57 28 PM


OpenRDTReaderWeb/firebase.json
```diff
@@ -1,12 +1,34 @@
...
+    },
+    "firestore": {
+      "port": 8080
+    },
+    "database": {
+      "port": 9000
+    },
+    "hosting": {
+      "port": 5000
+    }
```
ethantmanns, line 32: This is for local testing which I'm still figuring out

OpenRDTReaderWeb/firestore.rules
```diff
@@ -0,0 +1,16 @@
...
+
+    function isLoggedInUser(userUID) {
+      return request.auth != null && request.auth.uid == userUID;
+    }
+
+    match /users/{user}/{document=**} {
+      allow read, write: if isUnsetUser(user) || isLoggedInUser(user);
+    }
+  }
+}
```
ethantmanns, line 16: For now I just tested this in the web console

OpenRDTReaderWeb/firestore.rules
```diff
@@ -0,0 +1,16 @@
+rules_version = '2';
+service cloud.firestore {
+  match /databases/{db}/documents {
+    function isUnsetUser(userUID) {
+      return userUID == "UNSET_USER_ID";
+    }
```
ethantmanns, line 6: Need to find out if this is dangerous. I think any guest user could change the data of other guests' test runs.

OpenRDTReaderWeb/src/utils/AppConfig.ts
```diff
@@ -18,6 +18,8 @@ const MasterConfig: AppConfig = {
   staticPhotoMLRecognitionEnabled: true,
 
   landingPage: "NEWTESTRUN",
+
+  requireAuth: false,
```
ethantmanns, line 22: Probably going to rename this to "allowGuest"

OpenRDTReaderWeb/src/store/index.ts
```diff
@@ -83,7 +85,10 @@ function authReducer(state = initialAuthState, action: AuthAction): AuthState {
     case "AUTH_LOGGED_OUT":
       return {
         loading: false,
-        auth: null,
+        auth: {
+          uid: UNSET_USER_ID,
+          emailVerified: true,
```
ethantmanns, line 90: Not a fan of hardcoding this to a specific ID 🤔 

Also we mark email verified to true so that it doesn't get caught up in the verification flow, but probably need to revisit this in case we end up using `emailVerified` for other things

OpenRDTReaderWeb/src/components/TestRun/NewTestRun.tsx
```diff
@@ -1,25 +1,46 @@
...
+
+      if (!authUID && !config.requireAuth) {
+        userUID = UNSET_USER_ID;
+        const unsetUserExists = await firebase.doesUserExist(UNSET_USER_ID);
+        if (!unsetUserExists) {
+          await firebase.addUser(UNSET_USER_ID);
+        }
+      } else if (userUID === undefined) {
+        throw new Error("Auth is required");
+      }
```
ethantmanns, line 34: Wasn't sure if I should do this programatically like this. We could just have whatever deploy script we use generate this unset user instead...

OpenRDTReaderWeb/src/routes/routes.ts
```diff
@@ -46,17 +47,19 @@ function always(): boolean {
 }
 
 function loggedInVerified(auth: AuthInfo): boolean {
-  return !!(auth && auth.emailVerified);
+  return !!(auth && auth.emailVerified) && !auth.isGuest;
```
yrichard, line 50: It feels weird to check for isGuest here. Is there a chance auth.isGuest and auth.emailVerified would both be true?

OpenRDTReaderWeb/src/store/index.ts
yrichard: it feels like emailVerified should be false.... Can you point where this causes a problem?

OpenRDTReaderWeb/src/routes/routes.ts
yrichard: Ah I see below. What's happening if you set emailVerified to false for guests?

ethantmanns: In DashboardApp we redirect to verify email. I'm just going to change that logic to not do that for guest. Idk why I didn't think to do that before...

ethantmanns:
I'm marking this one as ready for review while I still work on getting the tests working

OpenRDTReaderWeb/firestore.rules
balderdash: That's correct. The way we often handle this in other projects is by using a uuid as the document ID, and only granting unauthenticated users `write` permission, not `read`. In theory the random document IDs should be enough if you don't provide the `list` firestore permission, but it's better if we can avoid giving out `read` permission to unauthenticated users at all.

OpenRDTReaderWeb/src/components/TestRun/NewTestRun.tsx
```diff
@@ -1,25 +1,46 @@
...
-      const newTestRunUid = await getFirebaseApp().addTestRunToUserForProfile({
+      let userUID = authUID;
+      const firebase = getFirebaseApp();
+
+      if (!authUID && config.allowGuestRuns) {
+        userUID = UNSET_USER_ID;
+        const unsetUserExists = await firebase.doesUserExist(UNSET_USER_ID);
+        if (!unsetUserExists) {
+          await firebase.addUser(UNSET_USER_ID);
+        }
```
balderdash, line 31: I don't think this is necessary. I'm pretty sure Firestore will let you create a document in a subcollection of a document that doesn't exist yet, and since we don't need to save anything on the guest user document itself we can just skip this whole check.

OpenRDTReaderWeb/firestore.rules
ethantmanns: @balderdash I think we actually need to end up reading in order to update the timestamps for all the steps. We know we have to do some re-design of the schema so maybe the best way forward is to keep the test run info and the personal info separate with separate rules?

PR #683, opened 2020-06-03
OpenRDT Web: Make Questions Removable

yrichard:
I am adding a few functionalities I need for the TestRun Editor.

- Ability to remove Questions in a Questionnaire:
It is done by adding the "unregisterQuestion" api, which will be called when the QuestionnaireInput unmounts. This will be useful when adding items in a list.

- getDefaultQuestionDefinition: 
Boiler plate definition for most questions.

- onlyNumberFormatter:
I will need this for some component that require a number prop.


OpenRDTReaderWeb/src/components/ui/Questionnaire/Questionnaire.tsx
```diff
@@ -116,6 +120,16 @@ const Questionnaire = (props: Questionnaire) => {
     [questions, questionUpdaters, onFormReady, submitEnabled]
   );
 
+  const unregisterQuestion = useCallback(
+    (question: Question) => {
+      delete questions[question.name];
```
ethantmanns, line 125: TIL https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete

yrichard: yeah, that's old school core JS right there. It's pretty cool :)

PR #684, opened 2020-06-03
OpenRDT Editor Draft

yrichard:
Adding a first rough draft for the test run editor, focusing on creating the page content first. I will polish the UI at a later time.

Editor.tsx: Entry point for the editor.

EditableTypes.ts: Declaring types. Component that can be exposed in the editor will need to define an interface extending EditableType.
EditableType object will describe explicitly which props are exposed in the UI, along with their "valueType", see EditableAsset, EditableParagraph, EditableTitle, EditableNextStepButton and EditableDeclarativeList.

AddContentDialog.tsx: Dialog to add a component in the page and set its props.

EditorForComponent.tsx: Generic editor for components. Basically, going through the props and creating a Questionnaire to edit them.

DeclarativeListEditor.tsx: A custom editor for the bullet list (allowing to add/remove items). It's not great looking but functional.

Screen Shot 2020-06-03 at 3 41 05 PM
Screen Shot 2020-06-03 at 3 41 20 PM
Screen Shot 2020-06-03 at 3 41 29 PM
Screen Shot 2020-06-03 at 3 41 55 PM
Screen Shot 2020-06-03 at 3 43 12 PM
Screen Shot 2020-06-03 at 3 43 26 PM


OpenRDTReaderWeb/src/components/ui/Asset.tsx
```diff
@@ -15,10 +18,60 @@ export interface AssetsImageProps extends ImgHTMLAttributes<HTMLImageElement> {
...
+      { name: "local", value: "local" },
+    ],
+  },
+  src: {
+    ...getDefaultQuestionDefinition(),
+    label: "src",
+    mandatory: true,
+    name: "src",
+    valueType: "string",
+  },
```
ethantmanns, line 64: Maybe for a later iteration, but to me it's a little unclear what might be expected here in local mode

yrichard: yeah, anyways, I'm thinking we should have a custom editor here that shows the files available for locally and server side and hide those details away.

PR #689, opened 2020-06-05
USC-941 redux photo upload

ethantmanns:
Putting redux between the client and Firebase for photo uploading actions


OpenRDTReaderWeb/src/store/storage.ts
```diff
@@ -0,0 +1,104 @@
...
+        uploading: true,
+      };
+    case "UPLOAD_COMPLETE":
+      return {
+        uploading: false,
+      };
+    default:
+      return state;
+  }
+}
```
ethantmanns, line 55: Nothing is really using this since we're using callbacks for everything. Probably will remove unless someone sees a good use case for these?

yrichard: Is this some type of pattern that usually comes with Redux? I don't have a strong opinion, but if it's unused, then yeah probably not needed.

OpenRDTReaderWeb/src/store/storage.ts
```diff
@@ -0,0 +1,104 @@
...
+
+export async function uploadPhoto(uploadParams: PhotoUploadParams) {
+  const store = getStore();
+  store.dispatch({
+    type: "UPLOAD_PHOTO",
+  });
```
yrichard, line 61: are we actually using those dispatch calls?

OpenRDTReaderWeb/src/components/FileUploader/PhotoPreview.tsx
```diff
@@ -5,6 +5,7 @@
 
 import { ROUTE_DEFINITIONS, getPath } from "routes/routes";
 import React, { useCallback, useEffect, useState } from "react";
+import { UploadStatus, updateTestRunPhoto, uploadPhoto } from "store/storage";
```
yrichard, line 8: so it seems we're calling the functions directly and bypassing Redux, right? I have nothing against this approach, but I'm not sure this is the Redux way. @balderdash, what do you think?

ethantmanns: Oh should it instead be more of calling the action and letting the reducer do the task then update the state?

yrichard: I really don't have enough experience with redux to know what's the best practice here. I'm fine with your code, but since we're going for full redux, maybe there's a better way... 
I've found this article which explains how to get something similar to work:
https://dev.to/devinekadeni/how-to-upload-multiple-file-with-progress-bar-reactjs-redux-and-expressjs-4hb3

balderdash: Yeah I'd vote we try going full redux, which means getting rid of photo upload component state. I.e. progress, uploadState, and the storageUrl should be stored in redux state, and use something like `const { progress, uploadState, storageUrl } = usePhotoUploadState()` in the render function to load it.

You generally wouldn't want to have the reducer handle the task, because the reducer is supposed to be a pure function. There are various ways to handle this, but a simple option is to use the action creators to trigger the side effect, e.g.

```
function uploadPhoto(params) {
  return dispatch => {
    dispatch({ type: 'UPLOAD_PHOTO', ... });
    getFirebaseApp().uploadPhoto({
      ...params,
      onFileUploadComplete: () => dispatch ({ type: 'UPLOAD_COMPLETE', ...}),
      <other callbacks>
    });
  }
}
```

Then in this component you would just call `dispatch(uploadPhoto(...))` without the callbacks, and just let the state updates come to you via redux.

(Note that this style uses `redux-thunk` middleware, which is easy to add: https://github.com/reduxjs/redux-thunk, but you could also just pass the dispatch function as a param to the action creator explicitly.)

mathewsb: Yeah, `redux-thunk` is exactly how I was planning to do this, until the NHS decided to rewrite the app and I had to work on other things :)

OpenRDTReaderWeb/src/components/FileUploader/PhotoPreview.tsx
```diff
@@ -60,106 +47,99 @@ export default (props: PhotoPreviewProps) => {
...
-    };
-
-    setProgress(0);
-    setUploadingState(UPLOADING_STATES.BACKGROUND);
-
-    uploadAsync();
+  const { uploadState, uploadStatus, storageUrl } = useSelector(
+    (state: RootState) => state.storage
+  );
+  const dispatch = useDispatch();
```
ethantmanns: There are hooks for these!

OpenRDTReaderWeb/src/components/FileUploader/PhotoPreview.tsx
```diff
@@ -60,106 +47,99 @@ export default (props: PhotoPreviewProps) => {
...
-    };
-
-    setProgress(0);
-    setUploadingState(UPLOADING_STATES.BACKGROUND);
-
-    uploadAsync();
+  const { uploadState, uploadStatus, storageUrl } = useSelector(
+    (state: RootState) => state.photoUpload
+  );
+  const dispatch = useDispatch();
```
ethantmanns, line 60: Redux hooks :)

PR #700, opened 2020-06-12
OpenRDT Web: HealthPulse config with submit address form.

yrichard:
I am adding the skeleton of the test run for HealthPulse, using a specific config. We will add the art and text when ready.
I've also added a quick form to get the address, but it doesn't save it at the moment.

https://localhost:3000/c/healthpulse/home


OpenRDTReaderWeb/src/routes/routes.ts
```diff
@@ -275,7 +275,6 @@ export const ROUTE_DEFINITIONS: { [key: string]: DashboardRoute } = {
 ROUTE_DEFINITIONS.LANDING = {
   ...ROUTE_DEFINITIONS.LOGIN,
   path: "/",
-  customerSpecific: false,
```
yrichard, line 278: @balderdash I'm not sure why that was needed. But I need to remove it or clicking on the logo in the nav bar doesn't redirect to the correct customer area.

balderdash: This created the root level landing page, I suspect if you remove this and try to load localhost:300/ you'll hit some kind of "page not found". Can you change the navbar link to go to `LOGIN` instead of `LANDING`?

yrichard: Oh right, thanks!

yrichard: I think HOME is a better one.

PR #703, opened 2020-06-16
Move app config to redux

balderdash:
Part 1 of app config loaded from the server. Still just using the hardcoded config, but now we load it into redux on load and everything else reads it from there.


OpenRDTReaderWeb/src/routes/routes.ts
```diff
@@ -61,15 +61,26 @@ function loggedOut(auth: AuthInfo) {
   return !auth || !!auth.isGuest;
 }
 
-const config = getAppConfig();
+function loggedInVerifiedOrGuest(auth: AuthInfo, config: AppConfig) {
+  return config.allowGuestRuns ? always() : loggedInVerified(auth);
+}
+
+export const PAGE_ACCESS_OPTIONS = {
```
yrichard, line 68: is this done so that we can move some of the routes definitions into a server side config?

OpenRDTReaderWeb/src/components/App/DashboardApp.tsx
```diff
@@ -54,10 +60,10 @@ const DashboardApp = () => {
                   key={route.path}
                   path={getPath(route, ":customerPrefix")}
                   render={({ location, match }) => {
-                    if (loading) {
+                    if (authLoading || !config) {
                       return <Helmet title={`Open RDT: Loading`} />;
```
yrichard, line 64: Since we're doing this, could it be fair to assume that all the components within the page using "useAppConfig" will get a valid config object? And if so, could we simplify the code by removing the `if (!config) {... blah ...}`.
This isn't blocking the diff, but rather just a question.

OpenRDTReaderWeb/src/components/FileUploader/RDTImagePreview.tsx
```diff
@@ -6,30 +6,32 @@ import React, { useEffect, useState } from "react";
...
 }
 
-const config = getAppConfig();
-
 export default (props: RDTImagePreviewProps) => {
   const [detectionData, setDetectionData] = useState<TopDetectionData>();
   const classes = useTestResultPhotoUploaderStyle();
   const { dataURI } = props;
+  const { config } = useAppConfig();
+  if (!config) {
```
yrichard, line 21: You shouldn't have a conditional statement before any hooks. https://reactjs.org/docs/hooks-rules.html

OpenRDTReaderWeb/src/store/index.ts
```diff
@@ -54,11 +43,20 @@ export function useShallowEqualSelector<TSelected>(
...
-  return reducer(state, action);
+export type RootState = ReturnType<typeof rootReducer>;
+export type AppThunk<ReturnType = void> = ThunkAction<
+  ReturnType,
+  RootState,
+  unknown,
+  redux.Action<string>
+>;
+export type AppDispatch = ThunkDispatch<RootState, any, Action>;
+export function useDispatch(): AppDispatch {
```
yrichard, line 60: Should we use a different name for this, so that we don't confuse it with reactRedux.useDispatch?

OpenRDTReaderWeb/src/components/App/DashboardApp.tsx
balderdash: Yeah that's quite a bit cleaner. I updated `useAppConfig()` to assert that the config is loaded, and removed the `if (!config)` checks.

OpenRDTReaderWeb/src/routes/routes.ts
balderdash: This is just so that we can define the routes before the config is loaded, I don't expect the routes themselves to be moved to the server side config.

